PR #2, opened 2019-04-06
ClearLinux Support

electrocucaracha:
This change is an effort to add support of ClearLinux Distribution.


vagrant/ClearLinux.yml
```diff
@@ -0,0 +1,15 @@
...
+# SPDX-license-identifier: Apache-2.0
+##############################################################################
+# Copyright (c) 2019 Intel Corporation
+# All rights reserved. This program and the accompanying materials
+# are made available under the terms of the Apache License, Version 2.0
+# which accompanies this distribution, and is available at
+# http://www.apache.org/licenses/LICENSE-2.0
+##############################################################################
+qat_driver_pkgs:
+  - linux-lts2018-dev
```
mythi, line 11: why this kernel?

vagrant/krd-vars.yml
```diff
@@ -8,8 +8,13 @@
 # http://www.apache.org/licenses/LICENSE-2.0
 ##############################################################################
 
-qat_driver_dest: "/tmp/qat"
-qat_driver_version: "1.7.l.4.3.0-00033"
-qat_driver_url: "https://01.org/sites/default/files/downloads/intelr-quickassist-technology/qat{{ qat_driver_version }}.tar.gz"
 # NOTE: Replace the `qat_envoy_dest` value with the current absolute path
 qat_envoy_dest: "/home/vagrant/kubernetes-qat-envoy/"
+qat_driver_version: "1.7.l.4.5.0-00034"
+qat_plugin_version: "1.7.l.4.3.0-00033"
```
mythi: these must be the same version

vagrant/_commons.sh
```diff
@@ -30,13 +27,24 @@ function install_k8s {
...
+        case ${ID,,} in
+            rhel|centos|fedora)
+                sudo yum install -y git
+            ;;
+            clear-linux-os)
+                sudo systemctl unmask docker.service
+            ;;
+        esac
+        #sudo git clone https://github.com/kubernetes-sigs/kubespray $kubespray_folder
+        sudo git clone https://github.com/electrocucaracha/kubespray $kubespray_folder
```
mythi: my expectation is when running on Clear Linux we use k8s what Clear linux ships in ```cloud-native-basic``` bundle. The bundle provides easy-to-use installation scripts in ```/usr/share/clr-k8s-examples```

vagrant/ClearLinux.yml
electrocucaracha: This vagrant image came with LTS version

vagrant/_commons.sh
electrocucaracha: AFAIK this PR does that https://github.com/kubernetes-sigs/kubespray/pull/3855/files

mythi: Does it work for you? A quick look tells me it installs ```containers-basic``` (and not ```cloud-native-basic```)

mythi: I did some checking and apparently kubespray always uses the upstream binaries and not what the OS provides (for cri-o, kubeadm/kubelet). It also does not give you kata-containers.

electrocucaracha: We can do those changes after Kubernetes is deployed. Eventually we can propose those changes to the Kubespray community

mythi: > We can do those changes after Kubernetes is deployed. 

I don't follow what changes? My proposal is for Clear install we use ```cloud-native-basic``` and the setup scripts it provides.

vagrant/installer.sh
```diff
@@ -25,8 +25,20 @@ chmod og-wx ~/.ssh/authorized_keys
...
-curl -sL https://bootstrap.pypa.io/get-pip.py | sudo python
-sudo -E pip install ansible
+if [ -e /etc/fstab ]; then
+    sudo sed -i '/ swap / s/^/#/' /etc/fstab
+fi
+# shellcheck disable=SC1091
+source /etc/os-release || source /usr/lib/os-release
+case ${ID,,} in
+    rhel|centos|fedora)
+        curl -sL https://bootstrap.pypa.io/get-pip.py | sudo python
```
ipuustin, line 40: Doesn't `pip` or `ansible` come packaged by the distros? We would get a known version and safer installation this way.

vagrant/sample_plugin.sh
```diff
@@ -0,0 +1,35 @@
...
+    popd
+fi
+if ! kubectl get secret --no-headers | grep -e envoy-tls-secret; then
+    openssl req -x509 -new -batch -nodes -subj '/CN=localhost' -keyout /tmp/key.pem -out /tmp/cert.pem
+    kubectl create secret tls envoy-tls-secret --cert /tmp/cert.pem --key /tmp/key.pem
+fi
+for resource in k8s_resources/*.yml; do
+    kubectl apply -f "$resource"
+done
+qat_svc=$(sudo service qat_service status | grep "There is .* QAT acceleration device(s) in the system:")
```
ipuustin: Parsing the service output seems bit fragile. Isn't there a better way to check if QAT is up?

vagrant/prechecks.sh
```diff
@@ -25,11 +25,14 @@ if ! sudo -n "true"; then
...
-        if [[ $ip =~ ^[0-9]+\.[0-9]+\.[0-9]+\.[0-9]+$  &&  $no_proxy != *"$ip"* ]]; then
+        if [[ $ip =~ ^[0-9]+\.[0-9]+\.[0-9]+\.[0-9]+$  &&  $NO_PROXY != *"$ip"* ]]; then
             echo "The $ip IP address is not defined in no_proxy env"
             exit 1
         fi
     done
 fi
+
+#echo "Making sure the system clock is accurate..."
+#sudo date -s "$(wget -qSO- --max-redirect=0 google.com 2>&1 | grep Date: | cut -d' ' -f5-8)Z"
```
ipuustin: It's bit unclear here whether the clock should be accurate or not. If yes, why comment this out?

vagrant/postchecks_qat_plugin.sh
```diff
@@ -15,33 +15,36 @@ set -o errexit
...
 
 echo "Verifying intel-qat2-plugin daemonset is available..."
 plugin_daemonset=$( kubectl get daemonset | grep intel-qat2-plugin)
-if [[ $plugin_daemonset ]]; then
-   if [[ $(echo $plugin_daemonset | awk '{print $2}') != $(echo $plugin_daemonset | awk '{print $6}') ]]; then
-      echo "Ther Intel QAT daemonset plugin is not available yet"
-      exit 1
-   fi
+if [[ -n $plugin_daemonset ]]; then
+    if [[ $(echo "$plugin_daemonset" | awk '{print $2}') != $(echo "$plugin_daemonset" | awk '{print $6}') ]]; then
```
ipuustin: This needs at least a comment (what are the fields which you are comparing)? Maybe something like this would also work:

    kubectl get pods -l app=intel-qat2-plugin -o json | jq '.items[] | {status: .status.phase}'

vagrant/postchecks_qat_driver.sh
```diff
@@ -16,10 +16,10 @@ echo "QAT driver validation"
 
 # Driver validation
 supported_dev="c6xx dh895xcc"
-qat_svc=$(sudo service qat_service status | grep "There is .* QAT acceleration device(s) in the system:")
+qat_svc=$(sudo systemctl status qat_service | grep "There is .* QAT acceleration device(s) in the system:")
```
ipuustin: It seems that `adf_ctl status` return value is zero if acceleration devices are present in the system. If you can verify this, there is no need to parse the (fragile) service logging output.

vagrant/sample_plugin.sh
electrocucaracha: The QAT service is up in Virtual Machines as well. Parsing the output can help us to skip the check for VMs.

vagrant/prechecks.sh
electrocucaracha: We have detected that this way to fix the timezone is not accurate, but we should have a way to do it. This is more like a placeholder for a proper solution.

vagrant/postchecks_qat_plugin.sh
electrocucaracha: Agree, I'm going to include something in the next commit.

vagrant/postchecks_qat_driver.sh
electrocucaracha: It returns zero if the command was executed properly:
```
[vagrant@localhost ~]$ sudo /usr/local/bin/adf_ctl status
Checking status of all devices.
There is 0 QAT acceleration device(s) in the system:
[vagrant@localhost ~]$ echo $?
0
```

vagrant/installer.sh
```diff
@@ -49,6 +49,19 @@ ansible-playbook -vvv -i ./inventory/hosts.ini configure-qat.yml | tee setup-qat
 
 # Kubernetes installation
 install_k8s
+if [[ "${CONTAINER_RUNTIME:-runc}" == "kata" ]]; then
+case ${ID,,} in
+    clear-linux-os)
+        sudo swupd bundle-add containers-virt
+        kubectl apply -f https://raw.githubusercontent.com/kata-containers/packaging/master/kata-deploy/kata-rbac.yaml
+        kubectl apply -f https://raw.githubusercontent.com/kata-containers/packaging/master/kata-deploy/kata-deploy.yaml
```
mythi: The pre-built kata artifacts are not sufficient for the QAT usage. We will need the QAT VF drivers in Kata rootfs and the kernel with certain ```CONFIG_``` options enabled.

vagrant/installer.sh
```diff
@@ -49,6 +49,19 @@ ansible-playbook -vvv -i ./inventory/hosts.ini configure-qat.yml | tee setup-qat
 
 # Kubernetes installation
 install_k8s
+if [[ "${CONTAINER_RUNTIME:-runc}" == "kata" ]]; then
```
mythi: It does not have to be either ```runc``` or ```kata``` at this point. In kubernetes the pod can choose which runtime it needs to run via ```RuntimeClassName```.  Or is this switch to say "no kata this time"? 

Kata is not ```clear-linux-os``` specific either so we should check with regardless of the OS.

vagrant/installer.sh
electrocucaracha: Got it, I'm going to include the Eric's instructions here. I've decided if should be better to have ansible playbook or bash script for this.

vagrant/installer.sh
electrocucaracha: You're right the only instruction that applies for CLR is the addition of the bundle, I'll fix it.

vagrant/prechecks.sh
ipuustin: Let's remove the commented out code and add a `TODO` item instead.

vagrant/postchecks_qat_driver.sh
ipuustin: Did you try the negative case too? If both work then this might be a cleaner way to implement the check.

ipuustin:
I see you force pushed the branch a few times. Please comment if the PR is ready for another review round or if there are still unresolved issues left.

electrocucaracha:
> I see you force pushed the branch a few times. Please comment if the PR is ready for another review round or if there are still unresolved issues left.

@ipuustin it's failing when Kata is selected as container runtime. I'll let you know when this can be ready to be reviewed

electrocucaracha:
@ipuustin @mythi I have completed the support for ClearLinux. I need to work on the instructions to don't use the pre-built Kata images, but in the meantime could you review this patch? I created the `test.sh` script for testing all the possible combinations

vagrant/krd-vars.yml
```diff
@@ -8,8 +8,34 @@
...
+intel_plugin_url:
+  kata:
+    img_name: "intel-qat-plugin"
+    deployment_files:
+      - "qat_plugin/qat_plugin.yaml"
+      - "qat_plugin/qat_plugin_default_configmap.yaml"
+    url: "https://github.com/intel/intel-device-plugins-for-kubernetes"
+    version: "b2100f2735f0a8b759d5a34f2ccc6f80f422d499"
+  runc:
+    img_name: "intel-qat2-plugin"
```
mythi: everything uses intel-qat-plugin now. with ```-mode kernel``` you get what qat2-plugin did. (see the deployment yamls in the upstream repo)

vagrant/installer.sh
```diff
@@ -25,8 +25,20 @@ chmod og-wx ~/.ssh/authorized_keys
 
 # Install dependencies
 sudo swapoff -a
-curl -sL https://bootstrap.pypa.io/get-pip.py | sudo python
-sudo -E pip install ansible
+if [ -e /etc/fstab ]; then
```
mythi, line 33: To disable swap, Clear linux instructs 
```
sudo systemctl mask $(sed -n -e 's#^/dev/\([0-9a-z]*\).*#dev-\1.swap#p' /proc/swaps) 2>/dev/null
sudo swapoff -a
```

vagrant/test.sh
```diff
@@ -0,0 +1,95 @@
...
+##############################################################################
+
+set -o pipefail
+set -o errexit
+set -o nounset
+
+function usage {
+    cat <<EOF
+usage: $0 -r <REPO_URL> -d <TEST_DIR> -b <BRANCH>
+Run functional tests in different screen sessions that verifiies functionality
```
mythi: nit: verifies

vagrant/_commons.sh
```diff
@@ -12,29 +12,131 @@ set -o pipefail
...
+    fi
+    if [ -n "$HTTPS_PROXY" ]; then
+        echo "[Service]" | sudo tee /etc/systemd/system/docker.service.d/https-proxy.conf
+        echo "Environment=\"HTTPS_PROXY=$HTTPS_PROXY\"" | sudo tee --append /etc/systemd/system/docker.service.d/https-proxy.conf
+    fi
+    if [ -n "$NO_PROXY" ]; then
+        echo "[Service]" | sudo tee /etc/systemd/system/docker.service.d/no-proxy.conf
+        echo "Environment=\"NO_PROXY=$NO_PROXY\"" | sudo tee --append /etc/systemd/system/docker.service.d/no-proxy.conf
+    fi
+    if [[ "${CONTAINER_RUNTIME:-runc}" == "kata" ]]; then
```
mythi: What's the use case for having kata as the default runtime for docker?

vagrant/_commons.sh
```diff
@@ -12,29 +12,131 @@ set -o pipefail
 set -o xtrace
 set -o nounset
 
+# Kubespray configuration values
+kubespray_folder=/opt/kubespray
+
+# uninstall_k8s() - Uninstall Kuberentes deployment
```
mythi: nit: Kubernetes

vagrant/_commons.sh
```diff
@@ -12,29 +12,131 @@ set -o pipefail
...
+        source /etc/os-release || source /usr/lib/os-release
+        case ${ID,,} in
+            rhel|centos|fedora)
+                sudo yum install -y git
+            ;;
+            clear-linux-os)
+                sudo swupd bundle-add git
+            ;;
+        esac
+        sudo git clone --depth 1 https://github.com/kubernetes-sigs/kubespray $kubespray_folder -b release-2.9
```
mythi: 2.10 is out

vagrant/_commons.sh
```diff
@@ -12,29 +12,131 @@ set -o pipefail
...
+        sudo chown -R "$USER" $kubespray_folder
+        sudo -E pip install -r $kubespray_folder/requirements.txt
+
+        echo "Kubespray configuration"
+        mkdir -p ./inventory/group_vars/
+        cp all.yml ./inventory/group_vars/all.yml
+        cp k8s-cluster.yml ./inventory/group_vars/k8s-cluster.yml
+        if [[ "${CONTAINER_MANAGER:-docker}" == "crio" ]]; then
+            _install_docker
```
mythi: How docker gets installed in case ```CONTAINER_MANAGER``` is not crio?

mythi (changes requested):
I've added some review comments. PTAL.

One bigger concern I have is with from where docker/crio get pulled and with what version for each distro when the install is run. And also what controls we have for that. To me it looks this a mix right now. Examples: 1) crio/docker have different versions for Clear vs CentOS. 2) We don't use kubernetes from Clear.

```RuntimeClass``` provides controls over runc/kata selection so we don't need logic for that.

**Comment by electrocucaracha** [comment]
> I've added some review comments. PTAL.
> 
> One bigger concern I have is with from where docker/crio get pulled and with what version for each distro when the install is run. And also what controls we have for that. To me it looks this a mix right now. Examples: 1) crio/docker have different versions for Clear vs CentOS. 2) We don't use kubernetes from Clear.
> 
> `RuntimeClass` provides controls over runc/kata selection so we don't need logic for that.

Regarding the control of the Docker/CRI-O version for Clearlinux they try to use the latest stable version, there are just few exceptions(when it's not compatible or buggy). For CentOS, I submitted a [PR](https://github.com/kubernetes-sigs/kubespray/pull/4582) for updating the CRI-O version, but I'm not sure about Docker.

File: vagrant/_commons.sh (hunk 17 above)
Comment (electrocucaracha) [reply vagrant/_commons.sh]: Docker is installed by Kubespray but I found an issue in Kubespray when CRI-O is selected and Helm enabled, apparently this combination doesn't install Docker and fails during the execution. CRI-O is considered as *experimental* feature in Kubespray.

File: vagrant/_commons.sh (hunk 14 above)
Comment (electrocucaracha) [reply vagrant/_commons.sh]: I declared `docker` as default value for `CONTAINER_MANAGER` and `runc` as `CONTAINER_RUNTIME`, from my perspective `Katacontainers` and `CRI-O` are still under development, isn't it?

Comment (mythi) [reply vagrant/_commons.sh]: I'd say when setting up k8s, we should use CRI-O/containerd as ```CONTAINER_MANAGER``` (CRI-O is OK to start with) but not Docker anymore.

With proper (and new enough) CRI runtime used, ```CONTAINER_RUNTIME``` in these scripts gets void because the pods use ```RuntimeClassName``` to choose their preferred runtime. There could be a boolean flag to check whether kata-deploy should be run.

File: vagrant/_commons.sh (hunk 17 above)
Comment (mythi) [reply vagrant/_commons.sh]: Does *experimental* mean the install may fail?

File: vagrant/_commons.sh (hunk 14 above)
Comment (mythi) [reply vagrant/_commons.sh]: To get this PR merged, we can keep ```CONTAINER_MANAGER``` as it is but let's not make separation in these scripts about kata vs runc.

Later let's change the scripts so that the k8s setup uses a CRI runtime.

File: vagrant/_commons.sh (hunk 16 above)
Comment (electrocucaracha) [reply vagrant/_commons.sh]: Fixed on cb7d95c

File: vagrant/_commons.sh (hunk 17 above)
Comment (electrocucaracha) [reply vagrant/_commons.sh]: I assume that it's the way to say that their CI is not covering that particular scenario but I'm seeing more efforts to increase the adoption in [v2.11](https://github.com/kubernetes-sigs/kubespray/milestone/10)

File: vagrant/_commons.sh (hunk 15 above)
Comment (electrocucaracha) [reply vagrant/_commons.sh]: Fixed on cb7d95c

File: vagrant/test.sh (hunk 13 above)
Comment (electrocucaracha) [reply vagrant/test.sh]: Fixed on cb7d95c

File: vagrant/_commons.sh (hunk 14 above)
Comment (electrocucaracha) [reply vagrant/_commons.sh]: Okay, let me create a commit that removes it

Comment (electrocucaracha) [reply vagrant/_commons.sh]: Removed on  d86583c

File: vagrant/installer.sh (hunk 12 above)
Comment (electrocucaracha) [reply vagrant/installer.sh:33]: Fixed on a7d3a7a

File: vagrant/_commons.sh [hunk 18]
```diff
@@ -90,7 +90,11 @@ function install_k8s {
             echo "download_container: false"
             echo "skip_downloads: false"
             } >> ./inventory/group_vars/all.yml
-            sed -i 's/^etcd_deployment_type: .*$/etcd_deployment_type: host/;s/^kubelet_deployment_type: .*$/kubelet_deployment_type: host/;s/^container_manager: .*$/container_manager: crio/' ./inventory/group_vars/k8s-cluster.yml
+            sed -i 's/^etcd_deployment_type: .*$/etcd_deployment_type: host/' ./inventory/group_vars/k8s-cluster.yml
+            sed -i 's/^kubelet_deployment_type: .*$/kubelet_deployment_type: host/' ./inventory/group_vars/k8s-cluster.yml
+            sed -i 's/^container_manager: .*$/container_manager: crio/' ./inventory/group_vars/k8s-cluster.yml
+            # TODO: https://github.com/kubernetes-sigs/kubespray/issues/4737
```
mythi, line 106: the workaround is to stop either docker or cri-o, depending on which don't need. Edit: I need to think how to resolve the case with running localhost docker registry with cri-o. I don't quite like that.

vagrant/_commons.sh
```diff
@@ -90,7 +90,11 @@ function install_k8s {
...
+            sed -i 's/^kube_version: .*$/kube_version: v1.13.5/' ./inventory/group_vars/k8s-cluster.yml
```
mythi, line 107: The commit says moving to 1.14.1 but here you set it to 1.13.5?

vagrant/_commons.sh
electrocucaracha: In an All-in-One setup you always need docker for building Envoy and Intel Plugin images. Moreover, [there is some errors during the execution of kubespray if Helm is enabled](https://github.com/kubernetes-sigs/kubespray/issues/4705). IMHO both services have to coexist.

vagrant/_commons.sh
electrocucaracha: The 1.14.1 is the default version for docker, the v1.13.5 version is for CRI-O  given that can't coexist with Docker service. I can remove it once [this](https://github.com/kubernetes/kubernetes/pull/76505) is released for v1.15

electrocucaracha: BTW, the idea of using +v1.14 is to consume the [kustomize](https://kubernetes.io/docs/concepts/overview/object-management-kubectl/kustomization/#) feature is it's integrated in kubectl

vagrant/_commons.sh
```diff
@@ -12,29 +12,125 @@ set -o pipefail
...
+    fi
+    if [ -n "$HTTPS_PROXY" ]; then
+        echo "[Service]" | sudo tee /etc/systemd/system/docker.service.d/https-proxy.conf
+        echo "Environment=\"HTTPS_PROXY=$HTTPS_PROXY\"" | sudo tee --append /etc/systemd/system/docker.service.d/https-proxy.conf
+    fi
+    if [ -n "$NO_PROXY" ]; then
+        echo "[Service]" | sudo tee /etc/systemd/system/docker.service.d/no-proxy.conf
+        echo "Environment=\"NO_PROXY=$NO_PROXY\"" | sudo tee --append /etc/systemd/system/docker.service.d/no-proxy.conf
+    fi
+    if [[ "${CONTAINER_MANAGER:-docker}" == "crio" ]]; then
```
mythi: let's drop this block: docker does not need to use kata-runtime in our setup.

vagrant/_commons.sh
mythi: building is different from running a local docker registry.  in case of cri-o or containerd, we can avoid using `dockerd` by doing something like `docker save <tag> | [podman|ctr] ...`  after building and then stop docker daemon.

Please look into avoiding local docker registry.

vagrant/_commons.sh
mythi: See my comment about `podman` and stopping `dockerd`. Can we workaround the `helm` issue the same way (i.e., only run docker when it's needed)? It's not nice but...

Remember the ultimate goal we have: these scripts must provide the exact same setup for the distros.

vagrant/postchecks_qat_driver.sh
```diff
@@ -16,22 +16,30 @@ echo "QAT driver validation"
...
-        fi
-        dev_number=$(echo "$device" | awk '{print $1}' | tr -d 'qat')
-        for ssl_key in NumProcesses NumberCyInstances NumberDcInstances; do
-            sudo crudini --get --existing "/etc/$dev_type$dev_number.conf" SSL $ssl_key > /dev/null
-        done
-    done <<< "$devices"
+    # shellcheck disable=SC1091
+    source /etc/os-release || source /usr/lib/os-release
+    case ${ID,,} in
+        rhel|centos|fedora)
```
mythi: no clearlinux?

vagrant/postchecks_qat_driver.sh
```diff
@@ -16,22 +16,30 @@ echo "QAT driver validation"
...
+            devices=$(sudo /etc/init.d/qat_service status | grep 'state: up')
+            while read -r device; do
+                dev_type=$(echo "$device" | awk '{print $4}' | tr -d ',')
+                if [[ "$supported_dev" != *"$dev_type"* ]]; then
+                    echo "The $dev_type device type is not supported by QAT envoy plugin"
+                    exit 1
+                fi
+                dev_number=$(echo "$device" | awk '{print $1}' | tr -d 'qat')
+                for ssl_key in NumProcesses NumberCyInstances NumberDcInstances; do
+                    sudo crudini --get --existing "/etc/$dev_type$dev_number.conf" SSL $ssl_key > /dev/null
```
mythi: this is makes sense for the "default" install but our setup cannot use the default configs.

vagrant/postchecks_qat_plugin.sh
```diff
@@ -15,33 +15,42 @@ set -o errexit
 echo "QAT plugin validation"
 
 echo "Ensuring that envoy-qat:devel docker image exists..."
-if [[ -z $(sudo docker images | grep envoy-qat | grep devel) ]]; then
+if ! sudo docker images | grep -E "envoy-qat.*devel"; then
     echo "envoy-qat:devel docker image doesn't exists"
     exit 1
 fi
+if [[ "${CONTAINER_MANAGER:-docker}" == "crio" ]]; then
+    if ! sudo crictl images | grep -E "intel-qat-plugin.*devel"; then
```
mythi, line 30: I believe `crictl` works for docker too (via dockershim)

mythi (changes requested):
Some more changes requested.

The key ask is to ensure we run the same k8s version on all installs and have the cri-o based setup not to depend on docker after building the images. The rationale for the latter is that for things like CI setup, we don't necessarily need to setup things like dashboard and can avoid using `helm` and docker after the images are prepared. This also makes the cluster setup closer to real environment.

electrocucaracha:
> Some more changes requested.
> 
> The key ask is to ensure we run the same k8s version on all installs and have the cri-o based setup not to depend on docker after building the images. The rationale for the latter is that for things like CI setup, we don't necessarily need to setup things like dashboard and can avoid using `helm` and docker after the images are prepared. This also makes the cluster setup closer to real environment.

Let me see if how can I implement that because in order to remove `helm` dependency it's necessary to install `tiller` only when `sample_plugin.sh` is executed and they have plans to deprecate it in v3.

vagrant/installer.sh
electrocucaracha: Well, this guarantees that requirement. BTW, Kubespray doesn't support [ansible v2.8 yet](https://github.com/kubernetes-sigs/kubespray/issues/4778)

vagrant/postchecks_qat_driver.sh
electrocucaracha: ClearLinux only supports Python3 and Crudini hasn't migrated to that version.

vagrant/postchecks_qat_driver.sh
electrocucaracha: Agree, but I think that we should improve the way that the device is configured. This check only determines is there is an entry in the configuration file.

vagrant/postchecks_qat_plugin.sh
electrocucaracha: Maybe, but this is only installed when CRI-O is selected as CONTAINER_MANAGER

vagrant/prechecks.sh
electrocucaracha: Fixed on c142bb4d15bc048d92867c73a7b714aa2f8f2c7a

vagrant/postchecks_qat_plugin.sh
electrocucaracha: Fixed on 4abb01037aa3d91e093f57f1a25732955d3c91fd

vagrant/_commons.sh
electrocucaracha: Fixed on aac3d332a1e7014fe02a9e2979e3c7d8c3b39e9e

vagrant/prechecks.sh
```diff
@@ -34,5 +34,4 @@ if [[ ${NO_PROXY+x} = "x" ]]; then
     done
 fi
 
-#echo "Making sure the system clock is accurate..."
-#sudo date -s "$(wget -qSO- --max-redirect=0 google.com 2>&1 | grep Date: | cut -d' ' -f5-8)Z"
+# TODO: Ensure that system cloud is accurated
```
mythi: system time?

vagrant/Vagrantfile
```diff
@@ -50,6 +50,8 @@ Vagrant.configure("2") do |config|
               echo "module.sig_unenforce" | sudo tee /etc/kernel/cmdline.d/allow-unsigned-modules.conf
               echo "intel_iommu=igfx_off" | sudo tee /etc/kernel/cmdline-removal.d/disable-iommu.conf
               clr-boot-manager update
+              sudo mkdir -p /etc/systemd/resolved.conf.d
+              printf "[Resolve]\nDNSSEC=false" | sudo tee /etc/systemd/resolved.conf.d/dnssec.conf
```
mythi, line 56: Is this necessary? Clear's systemd is built using `-Ddefault-dnssec=no`

vagrant/postchecks_qat_driver.sh
mythi: The device plugin when run with `-mode kernel` does the checking. If it's running OK and the node has QAT resources registered, I think we have sufficient testing in place (and it works on all distros).

I'd propose we drop `crudini` altogether

vagrant/postchecks_qat_plugin.sh
mythi: OK, makes sense.

Dockerfile.envoy
```diff
@@ -1,6 +1,6 @@
 FROM debian:sid as builder
 
-ADD qat1.7.l.4.3.0-00033.tar.gz .
+ADD qat1.7.l.4.4.0-00023.tar.gz .
```
juliocri, line 3: seems that version is not available to download: 

https://01.org/sites/default/files/downloads/intelr-quickassist-technology/qat1.7.l.4.4.0-00023.tar.gz

Dockerfile.envoy
```diff
@@ -1,6 +1,6 @@
 FROM debian:sid as builder
 
-ADD qat1.7.l.4.3.0-00033.tar.gz .
```
juliocri, line 3: This version seems to be availabe:

https://01.org/sites/default/files/downloads/intelr-quickassist-technology/qat1.7.l.4.3.0-00033.tar.gz

vagrant/Vagrantfile
electrocucaracha: Well, maybe it's related to the [Intel's network](https://github.com/AntonioMeireles/vagrant-guests-clearlinux/issues/3#issuecomment-494928154) but if I don't manually disable I get timeouts.

Dockerfile.envoy
electrocucaracha: @juliocri the url is different for that version https://01.org/sites/default/files/downloads/qat1.7.l.4.4.0-00023.tar.gz that's why I have commented [these lines ](https://github.com/electrocucaracha/kubernetes-qat-envoy/blob/clearlinux/vagrant/krd-vars.yml#L19-L20)

juliocri: yes :/ sorry about it, my bad.

vagrant/_commons.sh
electrocucaracha: @mythi I couldn't find in [podman](https://github.com/containers/libpod/blob/master/cmd/podman/build.go#L274-L302) a way to deal proxies like we do with [docker](https://github.com/electrocucaracha/kubernetes-qat-envoy/blob/master/vagrant/configure-qat-envoy.yml#L32-L42). Maybe we have to use `buildah`, but the version that is installed on CentOS is 1.7.1 which doesn't support [--http-proxy](https://github.com/containers/buildah/blob/master/docs/buildah-bud.md) argument.

vagrant/prechecks.sh
electrocucaracha: Fixed on 1122777783b94f60dd8be8f2e9ea7d5838c02e9e

vagrant/configure-qat-envoy.yml
```diff
@@ -62,32 +56,28 @@
...
+        - intel-qat-plugin
     - name: build QAT-accelerated envoy image
 # NOTE: docker_image module doesn't take proxy values defined in ~/.docker/config.json
 #      docker_image:
 #        dockerfile: Dockerfile.envoy
 #        path: "{{ qat_envoy_dest }}"
 #        name: envoy-qat
 #        tag: devel
-      command: docker image build -t envoy-qat:devel -f Dockerfile.envoy .
+      command: docker image build -t envoy-qat:devel --no-cache -f Dockerfile.envoy .
```
mythi: Adding `--pull` too is recommended to ensure we always use the latest base image

vagrant/_commons.sh
mythi: As suggested, you could do `docker save <tag> | podman ...` so that you don't need proxies.

mythi (commented):
Two minor comments.

@juliocri can you test the latest version and then let's get this merged if it works.

vagrant/_commons.sh
electrocucaracha: I tested that [change](https://gist.github.com/electrocucaracha/f11f9ed3fd32da6aa8c5c11d17d59152) but it didn't work for me.

mythi: I built `podman` on Clear and tested it:
```
$ sudo bin/podman load -i ~/intel-qat-plugin.tar
Getting image source signatures
Copying blob 48fed9e79c43 done
Copying blob 01092e5921c5 done
Copying blob 6189abe095d5 done
Copying blob 5c3fa42c4bcf done
Copying blob 87c747af6dc3 done
Copying config 6454a82ee2 done
Writing manifest to image destination
Storing signatures
Loaded image(s): localhost/intel-qat-plugin:devel
$ sudo crictl images
IMAGE                        TAG                 IMAGE ID            SIZE
localhost/intel-qat-plugin   devel               6454a82ee2c6d       36.5MB
```

I noticed `podman` was picky about having ` /etc/containers/policy.json` so I simply `sudo cp /usr/share/defaults/crio/policy.json /etc/containers/`

mythi:
@electrocucaracha one final thing. I discussed with @ipuustin about the multiple commits in this PR. Before merging, we want the commits squashed in logical changes so that after each commit the repo is still functional.

ipuustin:
@electrocucaracha: the idea is that we'll have a reasonable amount of usable commits in the repository. This means that the commits by themselves contain a certain feature, preferably with a commit message which explains it quite well. Especially the commits should not touch previous commits in the same PR -- just squash them together.

The "after each commit the repo is still functional" requirement is needed for use cases such as `git bisect`, which are used to find the exact commit when a bug was introduced to the repository.

vagrant/Vagrantfile
mythi: Does the vagrant image have the latest systemd from Clear? See https://github.com/clearlinux-pkgs/systemd/commit/f748d5dac428c99ef20a37edf38ffe7c2ccbb74e

vagrant/postchecks_qat_driver.sh
electrocucaracha: Fixed on 0b95b8d9aef3d69cfa01828c48025547929fff40 and 8d77c25655aa6c21cbdfe98ba3dd242e90cef454

vagrant/krd-vars.yml
electrocucaracha: Fixed on 8d77c25655aa6c21cbdfe98ba3dd242e90cef454

vagrant/installer.sh
electrocucaracha: Fixed on d86583c8d4e34dab3cc3c497870bc17b0a8e6481

vagrant/Vagrantfile
electrocucaracha: I had some [issues](https://github.com/AntonioMeireles/vagrant-guests-clearlinux/issues/3) before trying to use the [latest CLR image](https://app.vagrantup.com/AntonioMeireles/boxes/ClearLinux) but maybe I can give another try

vagrant/configure-qat-envoy.yml
electrocucaracha: Fixed on 9b359e0c100080f4af4d89756b2b6d4bb68de8d1

electrocucaracha:
@mythi and @ipuustin I tried to test every single commit that I made and I tried to isolate the changes to make clear their purpose but the latest version has evolved thanks to your comments, I can try to figure out how to squash some of them as long as this is the final version.

vagrant/Vagrantfile
mythi: Did it work?

electrocucaracha: Nope :( I'm going to keep trying to see if I can make it work.

mythi: Perhaps not spend too much time on it now. It can be dropped later when it works.

vagrant/installer.sh
mythi: Let's take those in a later PR.

mythi:
I don't think I have any more comments. `podman` still needs to be added but ready to take it later.

@electrocucaracha can you try out squashing some of the commits?

vagrant/_commons.sh
electrocucaracha: I meant the image is copied and listed in the catalog but the daemonset can't pull it to start it.

vagrant/installer.sh
electrocucaracha: Ok, I'm going to fix the installation of ansible to have better control

electrocucaracha: Fixed on a942dfa05db723a84f821c24109e0f14ae1b9683

vagrant/_commons.sh
mythi: It might be that localhost registry must be added to known registries in `crio.conf`

electrocucaracha:
> I don't think I have any more comments. `podman` still needs to be added but ready to take it later.
> 
> @electrocucaracha can you try out squashing some of the commits?

@mythi I created two branches in my repo, the [backup](https://github.com/electrocucaracha/kubernetes-qat-envoy/tree/backup) contains the original version with 43 commits ahead of intel:master and the new refactored [clearlinux](https://github.com/electrocucaracha/kubernetes-qat-envoy/tree/clearlinux) branch with 32 commits ahead.  I'm finalizing the tests and I'll let you know when I'm ready.

mythi:
@electrocucaracha thanks!

mythi:
@electrocucaracha I checked the latest changes and I'm still not happy with the commits. There are several "intermediate stages" still:

- enable cri-o, fix cri-o => it should be one 'enable cri-o' but no fixing afterwards
- enable clearlinux, disable iommu, drop containers-virt, dnnsec => just enable clearlinux
- add kata, drop kata for docker => add kata for cri-o only
- kubespray version bumps 2.9 and 2.10 => bump to 2.10

For future PRs, we also need to pay more attention to commit messages. "fix something" is not good unless the message also explains what was broken.

mythi (approved):
@electrocucaracha nice!

PR #3, opened 2019-04-09
Adding initial Docker image with Envoy+BoringSSL+QAT

juliocri:
Signed-off-by: Rivera Gonzalez, Julio C <julio.c.rivera.gonzalez@intel.com>

### DIFF
--- a/Dockerfile-boringssl-envoy
+++ b/Dockerfile-boringssl-envoy
@@ -1,0 +1,36 @@
+FROM debian:sid as builder
+
+RUN apt-get update
+
+RUN apt-get -y install git libtool \
+    cmake clang-format-7 automake ninja-build curl \
+    git build-essential wget libudev-dev libssl-dev openssl pkg-config autoconf autogen libtool \
+    openssl libssl-dev pkg-config zip g++ zlib1g-dev unzip python curl gnupg2 \
+    python3 gperf libcap-dev
+
+RUN curl -LO https://github.com/bazelbuild/bazel/releases/download/0.24.1/bazel-0.24.1-installer-linux-x86_64.sh
+RUN chmod +x bazel-0.24.1-installer-linux-x86_64.sh
+RUN ./bazel-0.24.1-installer-linux-x86_64.sh --user
+
+RUN git clone --recurse-submodules --single-branch --branch qat-provider https://github.com/ipuustin/envoy.git envoy
+
+# build Envoy
+RUN cd envoy && /root/.bazel/bin/bazel build --verbose_failures -c opt -s //source/exe:envoy-static
+
+# Start creating the target image
+FROM debian:sid-slim
+
+RUN apt-get update && apt-get -y install gettext-base
+
+COPY --from=builder /envoy/bazel-bin/source/exe/envoy-static /envoy-static
+# Defaulting env vars
+ENV QAT_SECTION_NAME SHIM
+ENV POLL_DELAY 2000000
+
+# Replacing env var values on envoy config and generating entrypoint script
+RUN echo "#!/bin/sh\n" \
+         "set -e\n" \
+         "/usr/bin/envsubst < /etc/envoy/config/envoy-conf.yaml > /tmp/envoy-conf.yaml\n" \
+         "/envoy-static -c /tmp/envoy-conf.yaml \$@\n" > /usr/local/bin/entrypoint.sh
+
+ENTRYPOINT ["/bin/sh", "/usr/local/bin/entrypoint.sh"]
--- a/deployments/boringssl-envoy-deployment.yaml
+++ b/deployments/boringssl-envoy-deployment.yaml
@@ -1,0 +1,124 @@
+apiVersion: v1
+kind: Service
+metadata:
+  name: helloenvoy
+  labels:
+    app: boringssl-envoy
+spec:
+  type: NodePort
+  ports:
+    - port: 9000
+      targetPort: 9000
+      protocol: TCP
+      name: https
+  selector:
+    app: boringssl-envoy
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: boringssl-envoy-config
+data:
+  # envoy-conf.yaml - Configmap supports ENV variables, such:
+  # (String) $QAT_SECTION_NAME="SHIM"
+  # (Numeric) $POLL_DELAY=2000000
+  # To change default values,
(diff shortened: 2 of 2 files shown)


Dockerfile-boringssl-envoy
```diff
@@ -0,0 +1,32 @@
...
+# Start creating the target image
+FROM debian:sid-slim
+
+RUN apt-get update && apt-get -y install gettext-base
+
+COPY --from=builder /envoy/bazel-bin/source/exe/envoy-static /envoy-static
+
+RUN echo "#!/bin/sh\n" \
+         "set -e\n" \
+         "/usr/bin/envsubst < /etc/envoy/config/envoy-conf.yaml > /tmp/envoy-conf.yaml\n" \
```
ipuustin, line 33: It would be great if there was a default value for every required environmental variable. For example, set `QAT_SECTION_NAME="SHIM"` if `QAT_SECTION_NAME` is not set.

deployments/boringssl-envoy-deployment.yaml
```diff
@@ -0,0 +1,117 @@
...
+            port_value: 9000
+        filter_chains:
+          tls_context:
+            common_tls_context:
+              private_key_operations:
+                provider_name: qat
+                typed_config:
+                  "@type": "type.googleapis.com/qat.QatPrivateKeyOperationsConfig"
+                  section_name: ${QAT_SECTION_NAME}
+                  poll_delay: 2000000
```
ipuustin: Maybe make `poll_delay` set by env variable too? Would be interesting to try different values in automatic perf testing.

Dockerfile-boringssl-envoy
```diff
@@ -0,0 +1,32 @@
...
+    cmake clang-format-7 automake ninja-build curl \
+    git build-essential wget libudev-dev libssl-dev openssl pkg-config autoconf autogen libtool \
+    openssl libssl-dev pkg-config zip g++ zlib1g-dev unzip python curl gnupg2 \
+    python3 gperf libcap-dev
+
+RUN curl -LO https://github.com/bazelbuild/bazel/releases/download/0.24.1/bazel-0.24.1-installer-linux-x86_64.sh
+RUN chmod +x bazel-0.24.1-installer-linux-x86_64.sh
+RUN ./bazel-0.24.1-installer-linux-x86_64.sh --user
+
+RUN git clone --recurse-submodules --single-branch --branch qat-provider https://github.com/ipuustin/envoy.git envoy
```
ipuustin, line 15: This will not be the final place for this work, but let's keep the URL for now.

deployments/boringssl-envoy-deployment.yaml
```diff
@@ -0,0 +1,117 @@
...
+        - name: envoy
+          image: envoy-boringssl-qat:devel
+          securityContext:
+            privileged: true
+          imagePullPolicy: IfNotPresent
+          args:
+            - "--cpuset-threads"
+          env:
+            - name: loglevel
+              value: debug
```
ipuustin: Do not run perf tests with this log level...

ipuustin (changes requested):

deployments/boringssl-envoy-deployment.yaml
ipuustin: And maybe add a comment explaining that Kubernetes is not filling these values, but they are expected to be processed in the container.

Dockerfile-boringssl-envoy
```diff
@@ -0,0 +1,32 @@
...
+RUN cd envoy && /root/.bazel/bin/bazel build --verbose_failures -c opt -s //source/exe:envoy-static
+
+# Start creating the target image
+FROM debian:sid-slim
+
+RUN apt-get update && apt-get -y install gettext-base
+
+COPY --from=builder /envoy/bazel-bin/source/exe/envoy-static /envoy-static
+
+RUN echo "#!/bin/sh\n" \
```
ipuustin, line 31: An alternative for this script might be to use an init container (https://kubernetes.io/docs/concepts/workloads/pods/init-containers/) to create the final configuration file from the template?

juliocri:
Hi @ipuustin All changes requested are done in a new commit: https://github.com/intel/kubernetes-qat-envoy/pull/3/commits/66196b92e3b41234912b422126b13b00efceb12f on same PR.

deployments/boringssl-envoy-deployment.yaml
```diff
@@ -108,7 +115,23 @@ spec:
...
+          image: busybox
+          volumeMounts:
+          - name: workdir
+            mountPath: "/usr/share"
+          command: ['/bin/sh', '-c']
+          args:
+            - "echo '#!/bin/sh\n
+            set -e\n
+            /usr/bin/envsubst < /etc/envoy/config/envoy-conf.yaml > /tmp/envoy-conf.yaml\n
+            /envoy-static -c /tmp/envoy-conf.yaml $@' > /usr/share/entrypoint.sh"
```
ipuustin: You shouldn't run Envoy from the initContainer; just create the configuration file in it and let the initContainer run to completition. And I'm not even sure if initContainer is a better solution than doing an entrypoint script for the regular container -- what do you think?

juliocri: @ipuustin This init container only do that, write the config file in a volume, and then is terminated, after init container died envoy container starts but now the entrypoint.sh file is loaded in the right path to be executed by the entrypoint specified in the dockerfile.  But as same as you, I'm not quite sure of this approach is better, init adds few time more in starting  (not so much since is busybox and that image is only 1.2MB)...  at least with the dockerfile once the images is built we are sure that the entrypoint will be there and will be working... I haven't see another interesting approach, so I think I will move back to the Dockerfile is OK??

ipuustin: Ok, I think I just read the diff wrong. Sorry about the noise. :-) If you think going back to the one container approach makes more sense, please go ahead.

juliocri:
@ipuustin I think all changes requested are done, excepting the replace for the final url of the git repo. do you think we should wait to have that final url or we can merge this meanwhile? (I think can be very helpfully having merged in order to start adding it to the current jenkins pipeline)

ipuustin:
@juliocri Ok. Note that there will be changes to the configuration file format too. We can fix them as we go forward now -- I guess the important thing now is to have some tests for this work.

PR #7, opened 2019-05-03
 update to qat1.7.l.4.5.0-00034.tar.gz 

rojkov:


Dockerfile.envoy
```diff
@@ -27,7 +27,7 @@ ENV ICP_ENV_DIR /quickassist/build_system/build_files/env_files
 ENV ICP_BUILDSYSTEM_PATH /quickassist/build_system
 ENV KERNEL_SOURCE_ROOT /tmp
 
-RUN  ./configure --enable-qat-uio
+RUN  ./configure --disable-qat-lkcf
```
ipuustin, line 30: Why is this option needed?

.gitmodules
```diff
@@ -8,8 +8,8 @@
 	branch = section_name_env
 [submodule "intel-device-plugins-for-kubernetes"]
 	path = intel-device-plugins-for-kubernetes
-	url = https://github.com/rojkov/intel-device-plugins-for-kubernetes.git
-	branch = qat2
+	url = https://github.com/intel/intel-device-plugins-for-kubernetes.git
+	branch = master
```
ipuustin: Is this change actually necessary for the previous PR #6 to work? If yes, we should separate it.

README.md
```diff
@@ -13,7 +13,7 @@ Clone this repository with submodules:
 Then go to the created directory and fetch the QAT driver:
 
     $ cd kubernetes-qat-envoy
-    $ wget https://01.org/sites/default/files/downloads/intelr-quickassist-technology/qat1.7.l.4.3.0-00033.tar.gz
+    $ wget https://01.org/sites/default/files/downloads/qat1.7.l.4.5.0-00034.tar.gz
```
ipuustin, line 16: Let's add a checksum check here...

ipuustin (changes requested):

.gitmodules
rojkov: it's not necessary since the needed commit is present in both repos, but submitted a separate PR anyway.

README.md
rojkov: added

Dockerfile.envoy
rojkov: With this option there's no need to deal with `[KERNEL_QAT]` sections in config files.
And `--enable-qat-uio` was removed from the `./configure` script.

electrocucaracha: IMHO, Shouldn't be better to document the supported versions and leave this more *generic*? There is no guarantee that this tarball is that specific version, it only relies on the existence of `qat1.7.l.4.5.0-00034.tar.gz`

I suggest to change to:

```
ADD qat.tar.gz .
```

And modify the instruction in the README.md

mythi: or how about ```ARG QAT_DRIVER_RELEASE="qat1.7.l.4.3.0-00033"``` which you can override using build args and then something like ```ADD $QAT_DRIVER_RELEASE.tar.gz```?

rojkov: This PR also updates the reference to device plugins' repo which is configured to work with qat1.7.l.4.5.0-00034.tar.gz too. If we make configuration more flexible it'll put additional cognitive load on end users.

Then the QAT team is creative enough to put QAT releases to different URL paths, so that configuring the file name alone is not enough. And since the DP loads the tarball automatically when building docker images we'll need to complicate the code.

I feel we need to unify the way how `Dockerfile.envoy` and the corresponding `Dockerfile` in DP get the tarball. But I'd do it in a separate PR.

ipuustin:
I noticed that we have now inconsistency in `Dockerfile.envoy` and the instructions in `README.md`. I think this PR is the best way to fix that. Could you rebase and retest, and then we can merge?

rojkov:
Rebased and tested the image still builds ok with the new `Dockerfile.envoy`.

PR #11, opened 2019-06-14
vagrant: enable Intel IOMMU to support QAT pass-through

devimc:
Enable Intel IOMMU if it not enable, it's needed to support QAT pass-through
in virtual machines

Signed-off-by: Julio Montes <julio.montes@intel.com>

### DIFF
--- a/vagrant/Vagrantfile
+++ b/vagrant/Vagrantfile
@@ -90,6 +90,14 @@ Vagrant.configure("2") do |config|
     v.management_network_address = "192.168.123.0/27"
     v.management_network_name = "qat-mgmt-net"
     v.random_hostname = true
+    # Intel Corporation QuickAssist Technology
+    qat_devices = `for i in 0434 0435 37c8 6f54 19e2; do lspci -d 8086:$i -m; done|awk '{print $1}'`
+    qat_devices.split("\n").each do |dev|
+      bus=dev.split(':')[0]
+      slot=dev.split(':')[1].split('.')[0]
+      function=dev.split(':')[1].split('.')[1]
+      v.pci :bus => "0x#{bus}", :slot => "0x#{slot}", :function => "0x#{function}"
+    end
   end
 
   if ENV['http_proxy'] != nil and ENV['https_proxy'] != nil
--- a/vagrant/setup.sh
+++ b/vagrant/setup.sh
@@ -11,6 +11,7 @@
 set -o nounset
 set -o pipefail
 
+local modify_host=0
 vagrant_version=2.2.4
 if ! vagrant version &>/dev/null; then
     enable_vagrant_install=true
@@ -22,18 +23,38 @@ fi
 
 function usage {
     cat <<EOF
-usage: $0 -p <PROVIDER>
+usage: $0 -p <PROVIDER> [options]
 Installation of vagrant and its dependencies in Linux OS
 
 Argument:
     -p  Vagrant provider
+
+Options:
+    -m  Modify the host to enable the required features to use QAT.
+        WARNING: System files are modified. The system should be rebooted manually.
 EOF
 }
 
-while getopts ":p:" OPTION; do
+function enable_iommu {
+    echo "Enabling Intel_iommu"
+    sudo sed -i "s|^GRUB_CMDLINE_LINUX\(.*\)\"|GRUB_CMDLINE_LINUX\1 intel_iommu=on\"|g" /etc/default/grub
+    grub_cfg=$(sudo readlink -e /etc/grub2.cfg)
+    if sudo [ -f "${grub_cfg}" ]; then
+        echo "Updating grub configuration file"
+        sudo grub2-mkconfig -o "${grub_cfg}"
+        echo "System reboot is required"
+        return
+    fi
+    echo "grub configuration file does not exist"
+}
+
+while getopts ":p:m" OPTION; do
     case $OPTION in
     p)
         provider=$OPTARG
+        ;;
+    m)
+        modify_host=1
         ;;
     \?)
         usage
@@ -150,6 +171,9 @@ case ${ID,,} in
         packages+=(nfs-utils nfs-utils-lib)
         ;;
     esac
+    if [ ${modify_host} == 1 ] &&  [ ! -d /sys/class/iommu/* ] ; then
+        enable_iommu
+    fi
     ;;
 
 esac


vagrant/setup.sh
```diff
@@ -30,6 +30,24 @@ Argument:
...
+function enable_iommu {
+    echo "Checking Intel iommu"
+    if ! find /sys/kernel/iommu_groups/*/devices/* > /dev/null 2>&1; then
+        echo "Enabling Intel_iommu"
+        sudo sed -i "s|^GRUB_CMDLINE_LINUX\(.*\)\"|GRUB_CMDLINE_LINUX\1 intel_iommu=on\"|g" /etc/default/grub
+        grub_cfg=$(sudo readlink -e /etc/grub2.cfg)
+        if sudo [ -f ${grub_cfg} ]; then
+            echo "Updating grub configuration file"
+            sudo grub2-mkconfig -o ${grub_cfg}
+			echo "System reboot is required"
```
electrocucaracha: @devimc this line uses tabs instead of spaces
```suggestion
            echo "System reboot is required"
```

vagrant/setup.sh
```diff
@@ -30,6 +30,24 @@ Argument:
 EOF
 }
 
+function enable_iommu {
+    echo "Checking Intel iommu"
+    if ! find /sys/kernel/iommu_groups/*/devices/* > /dev/null 2>&1; then
+        echo "Enabling Intel_iommu"
+        sudo sed -i "s|^GRUB_CMDLINE_LINUX\(.*\)\"|GRUB_CMDLINE_LINUX\1 intel_iommu=on\"|g" /etc/default/grub
+        grub_cfg=$(sudo readlink -e /etc/grub2.cfg)
+        if sudo [ -f ${grub_cfg} ]; then
```
electrocucaracha: Could you surround the `${grub_cfg}` var with strings to pass shellcheck issues

```
In setup.sh line 39:
        if sudo [ -f ${grub_cfg} ]; then
                     ^---------^ SC2086: Double quote to prevent globbing and word splitting.

Did you mean: 
        if sudo [ -f "${grub_cfg}" ]; then
```

vagrant/setup.sh
```diff
@@ -30,6 +30,24 @@ Argument:
...
+            echo "Updating grub configuration file"
+            sudo grub2-mkconfig -o ${grub_cfg}
```
electrocucaracha: ditto: line 39

```
In setup.sh line 41:
            sudo grub2-mkconfig -o ${grub_cfg}
                                   ^---------^ SC2086: Double quote to prevent globbing and word splitting.

Did you mean: 
            sudo grub2-mkconfig -o "${grub_cfg}"
```

vagrant/setup.sh
electrocucaracha: In the other hand, what about to add a reboot instruction here, in that way we can enforce it.

electrocucaracha:
\cc @mythi

devimc:
@electrocucaracha changes applied, thanks

electrocucaracha:
\lgtm

vagrant/Vagrantfile
```diff
@@ -90,6 +90,14 @@ Vagrant.configure("2") do |config|
     v.management_network_address = "192.168.123.0/27"
     v.management_network_name = "qat-mgmt-net"
     v.random_hostname = true
+    # Intel Corporation QuickAssist Technology
+    qat_devices = `lspci | grep -e Quick -e QAT | awk '{print $1}'`
```
mythi: `grep -e Quick -e QAT` is not reliable. It may not find the cards if the PCI IDs are not updated (which is the case e.g., on my system) or it may find things that are not "supported" by the driver release.

I'd suggest to be explicit about the the devices

```
DEVICES="0434 0435 37c8 6f54 19e2"
qat_devices = `for i in $DEVICES; do lspci -d 8086:$i -m; done|awk '{print $1}'`
```

Also, before doing pass-through, you need to ensure the device is not bound to any driver.

vagrant/setup.sh
```diff
@@ -30,6 +30,24 @@ Argument:
 EOF
 }
 
+function enable_iommu {
+    echo "Checking Intel iommu"
+    if ! find /sys/kernel/iommu_groups/*/devices/* > /dev/null 2>&1; then
```
mythi: we use `/sys/class/iommu` elsewhere (e.g., in NFD).

ipuustin:
I had problems using QAT PFs when IOMMU was enabled (PFs didn't work at all). Is this not a concern after this change anymore? Do the vagrant scripts use only VFs? Did you test if VFs work right on bare metal?

(It would be great if we can standardize on VFs, because then we could change the DP to only support VF resources, simplifying the codebase.)

vagrant/setup.sh
```diff
@@ -150,6 +168,7 @@ case ${ID,,} in
         packages+=(nfs-utils nfs-utils-lib)
         ;;
     esac
+    enable_iommu
```
mythi: how about making the call conditional. e.g., run `enable_iommu` only if it's not enabled and it needs to be enabled.

vagrant/setup.sh
```diff
@@ -30,6 +30,24 @@ Argument:
 EOF
 }
 
+function enable_iommu {
+    echo "Checking Intel iommu"
+    if ! find /sys/kernel/iommu_groups/*/devices/* > /dev/null 2>&1; then
+        echo "Enabling Intel_iommu"
+        sudo sed -i "s|^GRUB_CMDLINE_LINUX\(.*\)\"|GRUB_CMDLINE_LINUX\1 intel_iommu=on\"|g" /etc/default/grub
```
mythi: Are all distros grub based (still)? Clear isn't but it comes with IOMMU enabled by default.

mythi (changes requested):

mythi:
> I had problems using QAT PFs when IOMMU was enabled (PFs didn't work at all)

AFAICS, in @devimc's setup, you don't use the PF driver with IOMMU on the host. This assigns the whole device to a VM that sees it as a unique device

mythi:
@devimc one other question: is this feature virtualization provider specific (i.e., it only works on libvirt)?

vagrant/Vagrantfile
electrocucaracha: @mythi in our initial tests we detected that those devices which were attached by another VM were ignored if those instructions are executed. In the other hand, it should be great to attach one device per VM, did you know if there is a way to detect which devices are not bound?

electrocucaracha:
> I had problems using QAT PFs when IOMMU was enabled (PFs didn't work at all). Is this not a concern after this change anymore? Do the vagrant scripts use only VFs? Did you test if VFs work right on bare metal?
> 
> (It would be great if we can standardize on VFs, because then we could change the DP to only support VF resources, simplifying the codebase.)

@ipuustin The idea of the `setup.sh` script is initialize/configure all the dependencies, that includes IOMMU

vagrant/setup.sh
electrocucaracha: I think that the conditional is included into the function isn't it?

vagrant/setup.sh
electrocucaracha: The line 144 of this file specifies the use case only for rhel|centos|fedora distributions

vagrant/Vagrantfile
electrocucaracha: @mythi in our initial tests we detected that those devices which were attached by another VM were ignored if those instructions are executed. In the other hand, it should be great to attach one device per VM, did you know if there is a way to detect which devices are not bound?

vagrant/setup.sh
electrocucaracha: I think that the conditional is included into the function isn't it?

vagrant/setup.sh
electrocucaracha: This script was designed to be used enabling QAT needs.

mythi: I'm not sure I follow. I just suggested an alternate path to check whether IOMMU is enabled.

vagrant/Vagrantfile
mythi: It's probably that `/sys/bus/pci/devices/0000:<BDF>/driver` does not exist.

vagrant/setup.sh
mythi: yes it is. I was suggesting to change for better readability: if (reasons to enable) then call enable_iommu

vagrant/setup.sh
mythi: of course. I lost the track that this was for that case only. thanks!

vagrant/setup.sh
electrocucaracha: @mythi is that alternate path is not based on the distro used?

mythi: that path is the kernel interface to check IOMMU is present

ipuustin:
@devimc : if we make large changes to the host system (such as changing the kernel command line), we need to tell the user well what's going to happen before they run the script. See https://github.com/intel/kubernetes-qat-envoy/issues/14

devimc:
@mythi 
>  is this feature virtualization provider specific (i.e., it only works on libvirt)?

it works on any hypervisor that support VFIO, like QEMU and NEMU

@ipuustin 
>  if we make large changes to the host system (such as changing the kernel command line), we need to tell the user well what's going to happen before they run the script. See #14

I agree, those changes should be documented. I have added an option in the script to allow changes in the host.


changes applied, thanks

mythi (approved):

PR #12, opened 2019-06-14
Fix non-declared PROXY env vars

electrocucaracha:
The `set -o nounset` bash instruction is causing issues when the proxy variables are not declared in the *_commons.sh* file. This change fixes that scenario.

### DIFF
--- a/vagrant/_commons.sh
+++ b/vagrant/_commons.sh
@@ -11,6 +11,7 @@
 set -o pipefail
 set -o xtrace
 set -o nounset
+set -o errexit
 
 # Kubespray configuration values
 kubespray_folder=/opt/kubespray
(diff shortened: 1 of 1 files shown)


mythi (changes requested):
Just one comment about the commit message.

You have nicely described the problem in the "intro letter", it's better to add it to the commit too:

"The set -o nounset bash instruction is causing issues when the proxy variables are not declared in the _commons.sh file. This change fixes that scenario."

vagrant/_commons.sh
```diff
@@ -79,10 +79,10 @@ function install_k8s {
         esac
         sudo git clone --depth 1 https://github.com/kubernetes-sigs/kubespray $kubespray_folder -b $kubespray_version
         sudo chown -R "$USER" $kubespray_folder
-        pushd $kubespray_folder
+        pushd "$kubespray_folder"  || exit
```
ipuustin: Or would it be simpler to use `set -o errexit`?

electrocucaracha: This change is for fixing a shellcheck warning.

ipuustin: I mean that you wouldn't need to handle `cd` errors if you had `errexit` set. `shellcheck` will know that too. But that might of course affect the rest of the script logic (and the scripts using this utility lib).

electrocucaracha: @ipuustin gotcha, I though that `set -o errexit` was declared on that file but it's only declared in the `installer.sh` file. Anyway, I'm going to include it and remove the `exit` instructions.

mythi: remember to add a note about it in the commit msg too

askervin (commented):
This slightly changes the behavior when `HTTP_PROXY` is `null`, that is an empty string. The original version never sets `HTTP_PROXY` to `null` in Docker contexts, but the new version does.

If you'd prefer to keep the old behavior and only fix the issue caused by `set -o nounset`, you may consider either using `if [[ -n ${HTTP_PROXY:+x} ]]`, or assigning default values to proxy variables in the beginning of the script and leaving all `if`'s as they are:

```
HTTP_PROXY=${HTTP_PROXY-}
HTTPS_PROXY=${HTTPS_PROXY-}
NO_PROXY=${NO_PROXY-}
```

mythi:
@electrocucaracha we did some PR cleanup. is this still relevant for you?

electrocucaracha:
> @electrocucaracha we did some PR cleanup. is this still relevant for you?

Well, it's required for environments outside of a corporate network. I tried to use the @askervin suggestions  but they didn't worked.

ipuustin:
Vagrant is not used anymore so closing.

PR #13, opened 2019-06-14
jenkins: Add initial Jenkinsfile

chuyd:
This PR integrates a basic Jenkinsfile which will trigger the ci
this first Jenkinsfile runs a simple stage that verifies the
qat-envoy docker image can be built. More functionallity will be
added in future commits.

Signed-off-by: Jesus Ornelas Aguayo <jesus.ornelas.aguayo@intel.com>


Jenkinsfile
```diff
@@ -0,0 +1,23 @@
...
+  stages {
+    stage('Create container for QAT-accelerated Envoy') {
+      options {
+        timeout(time: 90, unit: "MINUTES")
+        retry(2)
+      }
+      steps {
+        sh "wget https://01.org/sites/default/files/downloads/qat${QAT_VERSION}.tar.gz"
+        sh "sha256sum qat${QAT_VERSION}.tar.gz | grep $QAT_SHA_256"
+        sh "docker image build -t envoy-qat:devel -f Dockerfile.envoy ."
```
devimc: @chuyd nice patch!.  If #15 lands first, it will simplify the things, only this line will be needed

electrocucaracha: BTW, the docker build instruction should include some [additional parameters](https://github.com/intel/kubernetes-qat-envoy/blob/master/vagrant/configure-qat-envoy.yml#L67)

```suggestion
        sh "docker image build -t envoy-qat:devel --no-cache --pull -f Dockerfile.envoy ."
```

ipuustin: Does Jenkins have any built-in support for docker commands? It feels to me that caching could be smarter than just a on/off toggle.

ipuustin: Since #15 is now merged, you could simplify the file as suggested by @devimc to make use of the refactored build scripts.

Jenkinsfile
```diff
@@ -0,0 +1,23 @@
...
+    REPO_URL="https://github.com/intel/kubernetes-qat-envoy"
+  }
+  stages {
+    stage('Create container for QAT-accelerated Envoy') {
+      options {
+        timeout(time: 90, unit: "MINUTES")
+        retry(2)
+      }
+      steps {
+        sh "wget https://01.org/sites/default/files/downloads/qat${QAT_VERSION}.tar.gz"
```
ipuustin: I think it's not guaranteed that this is the download path. Maybe the URL could be an environmental variable too.

ipuustin (changes requested):

Jenkinsfile
chuyd: @ipuustin I updated the Jenkinsfile to make use of the refactored build scripts.

mythi: @chuyd there's a slight misunderstanding. We want `docker build` for all images. We can skip `wget` and the checksum verification since the script does that.

mythi (approved):
@chuyd @juliocri looks good but please rebase as discussed.

juliocri:
@chuyd take a look on: https://github.com/chuyd/kubernetes-qat-envoy/pull/1

chuyd:
@juliocri @mythi :^^

mythi:
@chuyd the job is failing.

juliocri:
@mythi ~Also manually clone with submodules, or submodule update fails  fetching `envoy-openssl` with:~

```
error: Server does not allow request for unadvertised object cfb700ab4719e1a57061d91df244c093fdbd4edc~
Fetched in submodule path 'envoy-openssl', but it did not contain cfb700ab4719e1a57061d91df244c093fdbd4edc. Direct fetching of that commit failed.
```

~.gitmodule file is correct?~

mythi:
@juliocri OK, I see. Yes, it's broken. We are sharing the commit that points to PR too and that branch has been updated.

/cc @rojkov

mythi:
@juliocri @chuyd it works now

mythi:
obsolete. we plan to use github actions for this provides

PR #15, opened 2019-06-21
Dockerfile: reimplement QAT/envoy build process

devimc:
Implement script to factorize the code that generates the envoy images.
Build time is also improved.
 
Signed-off-by: Julio Montes <julio.montes@intel.com>

### DIFF
--- a/Dockerfile.clr.envoy
+++ b/Dockerfile.clr.envoy
@@ -1,63 +1,12 @@
 FROM clearlinux:base as builder
 
-ADD qat1.7.l.4.3.0-00033.tar.gz .
 ADD openssl openssl
 ADD QAT_Engine QAT_Engine
 ADD envoy envoy
+ADD scripts scripts
+ADD versions.yaml versions.yaml
 
-# Add bundles and deps
-RUN rm -rf /run/lock/clrtrust.lock
-RUN clrtrust generate && swupd update && swupd bundle-add os-core-dev
-RUN curl -LO https://github.com/bazelbuild/bazel/releases/download/0.19.2/bazel-0.19.2-installer-linux-x86_64.sh && \
-    chmod +x bazel-0.19.2-installer-linux-x86_64.sh && \
-    ./bazel-0.19.2-installer-linux-x86_64.sh --user
-
-# Build QAT library
-ENV ICP_ROOT /
-ENV ICP_BUILD_OUTPUT /build
-ENV ICP_ENV_DIR /quickassist/build_system/build_files/env_files
-ENV ICP_BUILDSYSTEM_PATH /quickassist/build_system
-ENV KERNEL_SOURCE_ROOT /tmp
-
-RUN sed -i -e 's/cmn_ko$//' quickassist/Makefile && \
-    ./configure --enable-qat-uio && \
-    make -f quickassist/Makefile user && \
-    make adf-ctl-all && \
-    install -m 755 build/libqat_s.so /usr/lib && \
-    install -m 755 build/libusdm_drv_s.so /usr/lib && \
-    install -m 755 build/adf_ctl /usr/bin && \
-    install -d /etc/ld.so.conf.d && \
-    echo /usr/lib > /etc/ld.so.conf.d/qat.conf && \
-    ldconfig
-
-# Configure OpenSSL
-RUN cd /openssl && ./config
-
-# Build QAT Engine
-ENV PERL5LIB /openssl
-RUN cd /QAT_Engine && \
-    ./autogen.sh && \
-    ./configure \
-    --with-qat_dir=/ \
-    --with-openssl_dir=/openssl \
-    --with-openssl_install_dir=/usr/lib64 \
-    --enable-upstream_driver \
-    --enable-usdm \
-    --with-qat_install_dir=/usr/lib \
-    --enable-openssl_install_build_arch_path && \
-    make && \
-    make install && \
-    ldconfig
-
-# Build Envoy
-RUN install -d /usr/local/lib64 && \
-    cd /usr/local/lib64 && \
-    ln -s /usr/lib64/libssl.so && \
-    ln -s /usr/lib64/libcrypto.so
-
-RUN cd /envoy && \
-    mkdir -p /.cache/bazel/_bazel_root/ && \
-    /.bazel/bin/bazel build //source/exe:envoy-static
+RUN bash -f scripts/build_qat_envoy.sh
 
 # Start creating the target image
 FROM clearlinux:base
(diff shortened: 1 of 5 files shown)


devimc:
@electrocucaracha @ipuustin @mythi  Added version file to have a better control over the versions of the components

scripts/build_qat_envoy.sh
```diff
@@ -0,0 +1,203 @@
...
+
+set -e
+
+if [ -f /etc/os-release ]; then
+	source /etc/os-release
+else
+	source /usr/lib/os-release
+fi
+
+readonly YQ_BIN="$(realpath ${PWD}/yq)"
```
ipuustin: Shellcheck reports a bunch of problems with this file, for example `${PWD}` lacking quotes. Could you run the command and see which ones could be fixed? You can annotate false positives.

scripts/build_qat_envoy.sh
```diff
@@ -0,0 +1,203 @@
...
+readonly SCRIPTS_DIR="$(realpath "$(dirname "$0")")"
+readonly VERSIONS_FILE="${SCRIPTS_DIR}/../versions.yaml"
+readonly OPENSSL_DIR="${SCRIPTS_DIR}/../openssl"
+readonly QAT_ENGINE_DIR="${SCRIPTS_DIR}/../QAT_Engine"
+readonly ENVOY_DIR="${SCRIPTS_DIR}/../envoy"
+readonly QAT_LIB_DIR="${SCRIPTS_DIR}/../QAT_Lib"
+
+source "${SCRIPTS_DIR}/lib.sh"
+
+install_yq() {
```
ipuustin: You need `yq` for parsing the YAML configuration, right? It would be better if you used some tool already present in the distributions to make the security footprint smaller. Otherwise you may need to implement things like checksum checking for the downloaded file.

scripts/build_qat_envoy.sh
```diff
@@ -0,0 +1,203 @@
...
+	esac
+
+	curl -L https://github.com/mikefarah/yq/releases/download/2.4.0/yq_linux_${yq_arch} -o ${YQ_BIN}
+	chmod +x ${YQ_BIN}
+}
+
+install_bazel() {
+	info "Download and install bazel"
+	local bazel_url="$(${YQ_BIN} r "${VERSIONS_FILE}" "tools.bazel.url")"
+	local bazel_version="$(${YQ_BIN} r "${VERSIONS_FILE}" "tools.bazel.version")"
```
ipuustin: I think it's a good idea to make the versions parametric. Different Envoy versions require different bazel versions to build. You could add the required bazel checksums to the configuration file to verify the downloads.

scripts/build_qat_envoy.sh
```diff
@@ -0,0 +1,203 @@
...
+	tar -xf qat.tz
+
+	sed -i -e 's/cmn_ko$//' quickassist/Makefile
+	export ICP_ROOT="${QAT_LIB_DIR}" ICP_BUILD_OUTPUT="${QAT_LIB_DIR}/build" \
+		   ICP_ENV_DIR="${QAT_LIB_DIR}/quickassist/build_system/build_files/env_files" \
+		   ICP_BUILDSYSTEM_PATH="${QAT_LIB_DIR}/quickassist/build_system" KERNEL_SOURCE_ROOT=/tmp
+
+	case $ID in
+		clear-linux*)
+			./configure --enable-qat-uio
```
ipuustin, line 105: Note that required command line options for `libqat` compilation differ from version to version.

scripts/build_qat_envoy.sh
```diff
@@ -0,0 +1,203 @@
...
+			;;
+		debian|ubuntu)
+			./configure --disable-qat-lkcf
+			;;
+	esac
+
+	make -j $(($(nproc)-1)) -f quickassist/Makefile user
```
ipuustin: What happens if I only have one processor core?

scripts/build_qat_envoy.sh
```diff
@@ -0,0 +1,203 @@
...
+			;;
+	esac
+
+	ldconfig
+	popd
+}
+
+build_envoy() {
+	pushd "${ENVOY_DIR}"
+	mkdir -p ~/.cache/bazel/_bazel_root/
```
ipuustin: Is this needed?

ipuustin (changes requested):

scripts/build_qat_envoy.sh
devimc: get a better CPU :laughing: - just kidding, good catch! :+1:  I'll fix it, thanks

scripts/build_qat_envoy.sh
devimc: oops!, I forgot to enable shellcheck in emacs, I'll fix all errors

scripts/build_qat_envoy.sh
devimc: sure thing!

scripts/build_qat_envoy.sh
devimc: ok, it sounds feasible

devimc:
@ipuustin changes applied, thanks

scripts/build_qat_envoy.sh
```diff
@@ -0,0 +1,201 @@
...
+	local qat_url
+	local qat_version
+
+	qat_url="$(versions_yaml "libraries.qat.url")"
+	qat_version="$(versions_yaml "libraries.qat.version")"
+
+	mkdir -p "${QAT_LIB_DIR}"
+	pushd "${QAT_LIB_DIR}"
+
+	curl -L "${qat_url}/qat${qat_version}.tar.gz" -o qat.tz
```
ipuustin: I think it should be `qat.tgz` even though it doesn't matter that much. You should do the SHA256 check here too.

scripts/lib.sh
```diff
@@ -0,0 +1,17 @@
+#!/usr/bin/env bash
```
ipuustin, line 1: You could have the license header in this file too.

Dockerfile.clr.envoy
```diff
@@ -1,63 +1,12 @@
 FROM clearlinux:base as builder
 
-ADD qat1.7.l.4.3.0-00033.tar.gz .
```
ipuustin, line 3: This seems to be a functional change (pushing the QAT version up for clear linux image). The container built fine though.

ipuustin:
Some small nits but otherwise LGTM. @rojkov can you do a final pass after the changes are in?

devimc:
@ipuustin fixed, thanks

rojkov:
Thanks! LGTM

PR #17, opened 2019-07-08
build envoy from upstream envoy-openssl repository

rojkov:
The new [envoy-openssl](https://github.com/envoyproxy/envoy-openssl) repo allows building Envoy of OpenSSL flavours without excessive patching (though currently the lua HTTP filter is disabled since it depends on BoringSSL).

PR #18, opened 2019-07-17
Update libqat version to 4.6.

ipuustin:
Make the whole repository to only use one QAT version.

@electrocucaracha, @devimc, can you take a look? @juliocri, can you run the CI for this PR?

### DIFF
--- a/Dockerfile-boringssl-envoy
+++ b/Dockerfile-boringssl-envoy
@@ -5,14 +5,14 @@ RUN apt-get update
 RUN apt-get -y install git libtool \
     cmake clang-format-7 automake ninja-build curl \
     git build-essential wget libudev-dev libssl-dev openssl pkg-config autoconf autogen libtool \
-    openssl libssl-dev pkg-config zip g++ zlib1g-dev unzip python curl gnupg2 \
+    openssl libssl-dev pkg-config zip g++ zlib1g-dev unzip python curl \
     python3 gperf libcap-dev
 
-RUN curl -LO https://github.com/bazelbuild/bazel/releases/download/0.24.1/bazel-0.24.1-installer-linux-x86_64.sh
-RUN chmod +x bazel-0.24.1-installer-linux-x86_64.sh
-RUN ./bazel-0.24.1-installer-linux-x86_64.sh --user
+RUN curl -LO https://github.com/bazelbuild/bazel/releases/download/0.28.0/bazel-0.28.0-installer-linux-x86_64.sh
+RUN chmod +x bazel-0.28.0-installer-linux-x86_64.sh
+RUN ./bazel-0.28.0-installer-linux-x86_64.sh --user
 
-RUN git clone --recurse-submodules --single-branch --branch qat-provider https://github.com/ipuustin/envoy.git envoy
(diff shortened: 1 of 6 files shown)


Dockerfile-boringssl-envoy
```diff
@@ -5,14 +5,14 @@ RUN apt-get update
...
-    openssl libssl-dev pkg-config zip g++ zlib1g-dev unzip python curl gnupg2 \
+    openssl libssl-dev pkg-config zip g++ zlib1g-dev unzip python curl \
     python3 gperf libcap-dev
 
-RUN curl -LO https://github.com/bazelbuild/bazel/releases/download/0.24.1/bazel-0.24.1-installer-linux-x86_64.sh
-RUN chmod +x bazel-0.24.1-installer-linux-x86_64.sh
-RUN ./bazel-0.24.1-installer-linux-x86_64.sh --user
+RUN curl -LO https://github.com/bazelbuild/bazel/releases/download/0.28.0/bazel-0.28.0-installer-linux-x86_64.sh
+RUN chmod +x bazel-0.28.0-installer-linux-x86_64.sh
+RUN ./bazel-0.28.0-installer-linux-x86_64.sh --user
```
devimc, line 13: we have a different version of bazel in the `versions.yaml`. Should we update it?

ipuustin: We can update Bazel version everywhere when #17 lands and we can start tracking the Envoy master branch in every image. Let's wait until then. Only the BoringSSL image requires a recent Bazel because it is using a very recent Envoy version.

devimc: got it, thanks for explaining

devimc (approved):

versions.yaml
```diff
@@ -19,5 +19,5 @@ libraries:
 
   qat:
     url: "https://01.org/sites/default/files/downloads"
-    version: "1.7.l.4.5.0-00034"
-    sha256: "c42a3afc1a5c76d441eaca8b97dc1f9ee64939ec001539ee1a2f3b39b7543c8e"
+    version: "1.7.l.4.6.0-00025"
```
juliocri, line 22: @ipuustin when the driver is configured `./configure --flags` in Docker files or ansible files are called with `--enable-qat-uio`, but in this version that flag seems to be removed, a WARNING is shown, do you think it will be better if that flag is removed?

* https://github.com/intel/kubernetes-qat-envoy/blob/master/scripts/build_qat_envoy.sh#L105
* https://github.com/intel/kubernetes-qat-envoy/blob/master/vagrant/configure-qat.yml#L70

ipuustin: Good catch! Those flags are not used anymore in qat compilation. I added a patch which removes them.

juliocri (approved):
Latest CI run with with latest commit was a success: http://k8s-ci.intel.com:9090/blue/organizations/jenkins/kubernetes-qat-envoy/detail/test/14/pipeline/36

LGTM!

mythi:
@ipuustin thanks for the PR!

The version needs to be updated in quite many places still. Let me check if that can be changed to something where the version is defined in one place only and perhaps with a user override option.

PR #19, opened 2019-07-22
Update deployment config to current private key provider API

ipuustin:
@juliocri : please take a look.

### DIFF
--- a/deployments/boringssl-envoy-deployment.yaml
+++ b/deployments/boringssl-envoy-deployment.yaml
@@ -34,15 +34,15 @@ data:
         filter_chains:
           tls_context:
             common_tls_context:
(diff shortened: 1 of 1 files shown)


juliocri:
@ipuustin The env var is now set :), but the `fw_counters` are not increased yet, do you think is because the `intel-qat-plugin` is required to be updated to latest qat driver release? I see is still using the previous one: qat1.7.l.4.5.0-00034.tar.gz

ipuustin:
@juliocri I made one more necessary change to the config file (the `private_key` field is now owned by the QAT provider). Can you check if the tests pass now?

ipuustin:
@juliocri : one more fix, please try again.

PR #21, opened 2019-07-24
Drop k6

mythi:


juliocri (commented):
Looks good, doing the same for CI.

.gitmodules
```diff
@@ -2,19 +2,7 @@
 	path = envoy
 	url = https://github.com/ipuustin/envoy.git
 	branch = openssl-container
-[submodule "QAT_Engine"]
```
ipuustin, line 5: Good catch! I had forgotten `.gitmodules`.

README.md
```diff
@@ -63,23 +63,17 @@ Get the NodePort:
...
 
-In order to run benchmarks with k6 load testing tool, first create a container for a version which has `TLS_ECDHE_RSA_WITH_AES_128_CBC_SHA256` support enabled:
-
-    $ cd k6
-    # docker image build -t loadimpact/k6:custom -f Dockerfile .
-    $ cd ..
-
 Edit the `tests/k6-testing-config-docker.js` file to set the test parameters.  You can among other things select the cipher suite in the file. At least replace the port `9000` in the URL with the NodePort value.  Then run the benchmark:
 
-    # docker run --net=host -i loadimpact/k6:custom run --vus 10 --duration 20s -< tests/k6-testing-config-docker.js
```
ipuustin, line 82: Would be awesome if we could use some other version than "master" so that an upstream change couldn't break our testing, but we can do that later when a new K6 release comes out.

ipuustin (approved):
Looks good to me!

.gitmodules
mythi: I used `git rm <submodule>` in ba8774e and it updated .gitmodules too, FYI

README.md
mythi: Agree. I've subscribed for release notifications in loadimpact/k6 github so that we can change once there's a new release.

PR #16, opened 2019-06-22
Jenkins e2e test and perf using scripts in vagrant dir.

juliocri:

PR #22, opened 2019-08-07
Simplify openssl configuration by moving openssl.cnf to ConfigMap

rojkov:


Dockerfile.clr.envoy
```diff
@@ -23,9 +22,8 @@ RUN ln -s /usr/lib/libqat.so /usr/lib/libqat.so.0 && \
     install -d /etc/ld.so.conf.d && \
     install -d /etc/ssl && \
     echo /usr/lib > /etc/ld.so.conf.d/qat.conf && \
-    ldconfig && \
-    mv openssl.cnf /etc/ssl
+    ldconfig
 
 STOPSIGNAL SIGTERM
 
-ENTRYPOINT ["/envoy-static", "-c /etc/envoy/config/envoy-conf.yaml"]
```
ipuustin, line 31: Use rather `--cpuset-threads` parameter and control the cpuset size from Docker/K8s?

rojkov: Thanks! Updated.

ipuustin: Hmm, after some more investigation, it seems that in order for this to work the k8s CPU manager would need to be using the `static` policy, because otherwise the container cpuset isn't updated. The `static` policy isn't enabled by default and the `none` policy just uses the CFS quota.

I think that either we need to update the documentation saying that the CPU manager would need to be enabled or otherwise we'll need to manually sync the `--concurrency` value and cpu request value in the `yaml` file (instead of using `--cpuset-threads`). What do you think?

mythi: My preference would be that we add a note about CPU manager static policy and check the sample deployments get CPUs from the static pool.

FWIW, PR #16 has enablers for setting up a cluster with CPU manager. Also, our "cloud-native-setup" on Clear does that by default (a `kubeadm.yaml` config is provided).

rojkov:
Added a note about the static policy to `README.md`.

ipuustin (approved):
LGTM

mythi (approved):
great!

PR #23, opened 2019-08-12
Upgrade ClearLinux vagrant version

electrocucaracha:
The ClearLinux vagrant box used for this project hasn't been upgrade in a while so makes sense to have something more close to whatever that is available today. This patch pretends to update that image keeping version backwards compatibility.

### DIFF
--- a/vagrant/Vagrantfile
+++ b/vagrant/Vagrantfile
@@ -11,11 +11,11 @@
 box = {
   :virtualbox => {
     :centos => { :name => 'generic/centos7', :version=> '1.9.2' },
-    :clearlinux => { :name => 'AntonioMeireles/ClearLinux', :version=> '28510' }
+    :clearlinux => { :name => 'AntonioMeireles/ClearLinux', :version=> '30800' }
   },
   :libvirt => {
     :centos => { :name => 'centos/7', :version=> '1901.01' },
-    :clearlinux => { :name => 'AntonioMeireles/ClearLinux', :version=> '28510' }
+    :clearlinux => { :name => 'AntonioMeireles/ClearLinux', :version=> '30800' }
   }
 }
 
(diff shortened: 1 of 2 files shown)


vagrant/Vagrantfile
```diff
@@ -37,14 +37,21 @@ puts "[INFO] Shared folder: #{vagrant_root}"
 puts "[INFO] Linux Distro: #{distro}"
 puts "[INFO] Container manager: #{container_manager}"
 
+if distro == :clearlinux and Integer(box[:libvirt][distro][:version]) >= 28510
```
mythi: If this file takes care of pulling the right version, the backwards compatibility check is not needed.

mythi (commented):
Will this update help us to get rid of that dnssec setting that was needed earlier?

vagrant/Vagrantfile
electrocucaracha: Well, there is only one version for the OVMF.fd file so this can used for all CLR versions. The author of the CLR vagrant box [decided](https://github.com/AntonioMeireles/ClearLinux-packer/commit/005fcec22038866b9beba6fab4fab6a135c31b3f) to consume the host file.

I didn't check it but maybe the point of inflection happens around CLR [v29100](https://app.vagrantup.com/AntonioMeireles/boxes/ClearLinux/versions/29100)

mythi: > I didn't check it but maybe the point of inflection happens around CLR v29100

My comment was that the distro version check is not needed because you're also upgrading the box version. Or am I missing something?

electrocucaracha: Yes, you're right so can I assume that you don't need that backwards capability? I can change the code to remove that conditional.

mythi: Yes, that's my thinking. thanks!

vagrant/Vagrantfile
```diff
@@ -37,14 +37,21 @@ puts "[INFO] Shared folder: #{vagrant_root}"
...
+  File.exists?("/usr/share/qemu/OVMF.fd") ? loader = "/usr/share/qemu/OVMF.fd" : loader = File.join(File.dirname(__FILE__), "OVMF.fd")
+  if not File.exists?(loader)
+    system('curl -O https://download.clearlinux.org/image/OVMF.fd')
```
ipuustin: Shouldn't you check if the download succeeds? And does ruby have a built-in way to do HTTPs requests?

electrocucaracha: @ipuustin according to the Ruby's documentation the [system functio](https://ruby-doc.org/docs/ruby-doc-bundle/Manual/man-1.4/function.html#system)n returns false in case of failure so I can modified as well. Otherwise using a Ruby function I have to manage proxy values but it can be an alternative.

PR #24, opened 2019-08-13
Fix kubespray CRI-O template

electrocucaracha:
The template used for configure in CRI-O is pointing to the latest version which has variables not defined. This change fix the version of the patch used.

### DIFF
--- a/vagrant/_commons.sh
+++ b/vagrant/_commons.sh
@@ -94,7 +94,7 @@ function install_k8s {
                     sudo yum install -y wget
                 ;;
             esac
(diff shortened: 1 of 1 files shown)


mythi:
thanks

the template is still lagging behind. For example, it does not have settings for `RuntimeClass`. Is that reported in kubespray?

electrocucaracha:
> thanks
> 
> the template is still lagging behind. For example, it does not have settings for `RuntimeClass`. Is that reported in kubespray?

I have submitted a [PR](https://github.com/kubernetes-sigs/kubespray/pull/4582) that updates the CRI-O version and the configuration template

mythi (approved):

PR #29, opened 2019-10-31
Less aggresive number of bazel jobs

electrocucaracha:
This change reduces the number of bazel jobs by half to avoid building failures in environment with limited resources.

### DIFF
--- a/scripts/build_qat_envoy.sh
+++ b/scripts/build_qat_envoy.sh
@@ -197,11 +197,11 @@ build_envoy() {
 	pushd "${ENVOY_DIR}"
 	case $ID in
 		clear-linux*)
-		    CXXFLAGS="-Wno-error=stringop-truncation -Wno-error=redundant-move -DENVOY_SSL_VERSION=\\\"OpenSSL\\\"" ~/.bazel/bin/bazel build -j "$(jobs)" -c opt //:envoy --define boringssl=disabled
+		    CXXFLAGS="-Wno-error=stringop-truncation -Wno-error=redundant-move -DENVOY_SSL_VERSION=\\\"OpenSSL\\\"" ~/.bazel/bin/bazel build -j "HOST_CPUS*.5" -c opt //:envoy --define boringssl=disabled
 		    ;;
 
 		debian|ubuntu)
-		    CXXFLAGS="-Wno-error=stringop-truncation -Wno-error=redundant-move -DENVOY_SSL_VERSION=\\\"OpenSSL\\\"" ~/.bazel/bin/bazel build -j "$(jobs)" -c opt //:envoy --define boringssl=disabled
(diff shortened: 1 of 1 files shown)


mythi (changes requested):
@electrocucaracha I'd still like to understand better what those "some cases" are when it fails because this might just be a dirty workaround affecting everybody. The `jobs()` override could also be a `--build-arg` setting...

electrocucaracha:
> @electrocucaracha I'd still like to understand better what those "some cases" are when it fails because this might just be a dirty workaround affecting everybody. The `jobs()` override could also be a `--build-arg` setting...

@mythi you're right, that's ambiguous. My case is presented when I tried to build the envoy image from the worker node where processes like kubelet and kube-proxy require cpu cycles for their execution. Using a building argument is good idea but I'd prefer to have a conservative lower number as default option and give the possibility to increase it instead of having a higher default value.

mythi:
> have a conservative lower number as default

@electrocucaracha The default is to automatically scale to what `nproc` gives. Do you have enough memory?

electrocucaracha:
I just

> > have a conservative lower number as default
> 
> @electrocucaracha The default is to automatically scale to what `nproc` gives. Do you have enough memory?

I just checked and it seems a little low

```bash
vagrant@compute01:~$ free -g
              total        used        free      shared  buff/cache   available
Mem:             15           0           2           0          12          14
Swap:             0           0           0
```
```bash
vagrant@compute02:~$ free -g
              total        used        free      shared  buff/cache   available
Mem:             15           0           4           0          10          14
Swap:             0           0           0
```
This is part of the output that I get for the **compute01** worker node
```yaml
Capacity:
 cpu:                              24
 ephemeral-storage:                19478204Ki
 hugepages-1Gi:                    0
 hugepages-2Mi:                    0
 intel.com/intel_sriov_netdevice:  8
 memory:                           16429776Ki
 pods:                             110
 qat.intel.com/cy1_dc0:            24
 qat.intel.com/cy6_dc2:            1
Allocatable:
 cpu:                              23900m
 ephemeral-storage:                17951112777
 hugepages-1Gi:                    0
 hugepages-2Mi:                    0
 intel.com/intel_sriov_netdevice:  8
 memory:                           16077376Ki
 pods:                             110
 qat.intel.com/cy1_dc0:            24
 qat.intel.com/cy6_dc2:            1
```

ipuustin:
I'm on the opinion that the build should use `nproc` cores to make the build in the development and CI case as fast as possible. The builder can already restrict the resources which Docker build can use. For example, `docker build --cpu-quota=50000 ...` might do what you wish. Having a few extra threads isn't a problem unless you are running out of memory.

mythi:
not needed for now

PR #37, opened 2019-12-12
move to upstream envoy-openssl and update bazel

mythi:
Depends on envoyproxy/envoy-openssl/pull/10

### DIFF
--- a/scripts/build_qat_envoy.sh
+++ b/scripts/build_qat_envoy.sh
@@ -197,7 +197,7 @@ build_envoy() {
 	pushd "${ENVOY_DIR}"
 	case $ID in
 		clear-linux*|debian|ubuntu)
-		    CXXFLAGS="-DENVOY_SSL_VERSION=\\\"OpenSSL\\\"" ~/.bazel/bin/bazel build -j "$(jobs)" -c opt //:envoy --define boringssl=disabled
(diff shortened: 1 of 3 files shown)


scripts/build_qat_envoy.sh
```diff
@@ -197,7 +197,7 @@ build_envoy() {
 	pushd "${ENVOY_DIR}"
 	case $ID in
 		clear-linux*|debian|ubuntu)
-		    CXXFLAGS="-DENVOY_SSL_VERSION=\\\"OpenSSL\\\"" ~/.bazel/bin/bazel build -j "$(jobs)" -c opt //:envoy --define boringssl=disabled
```
ipuustin, line 200: I think that since envoy-openssl targets the latest Envoy, you'll have to bump bazel version too.

mythi:
@ipuustin @rojkov this is ready now

rojkov (approved):

mythi:
```
$cat envoy/.bazelversion
1.2.0
$
```

ipuustin:
Looks like the only difference between bazel 1.2.0 and 1.2.1 is a MacOS compatibility change. Since this compiles fine I think it's safe to merge.

mythi:
agree, thanks

PR #41, opened 2020-01-31
Update the BoringSSL QAT provider tree to latest.

ipuustin:
The biggest change is moving to use verson-managed API objects in Envoy. Also start using tags for version management. Do not use Python 2 in Bazel - the compilation succeeds with Python 3. Fix a trivial problem in the README.

### DIFF
--- a/Dockerfile.clr.boringssl.envoy
+++ b/Dockerfile.clr.boringssl.envoy
@@ -24,20 +24,20 @@ RUN mkdir /install_root \
     --no-boot-update \
     && rm -rf /install_root/var/lib/swupd/*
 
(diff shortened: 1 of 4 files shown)


rojkov (approved):
Looks good to me.

But I didn't fully understand the commit message: I see python2 is still installed in the build container.

ipuustin:
Wait a bit, I'll do one API renaming still.

ipuustin:
Changed the naming conventions a bit. Also fixed the explicit dep on python2 (I think both python2 and python3 end up being installed in the build container due to depedencies).

PR #42, opened 2020-03-18
add qatzip compression http filter

rojkov:
* QAT is linked dynamically to host's libqat_s.so and libusdm_drv_s.so since it's used also by QAT_Engine;
* QATZip is linked statically.


envoy_qatzip/compressor/qatzip_compressor_impl.cc
```diff
@@ -0,0 +1,101 @@
...
+  RELEASE_ASSERT(status == QZ_OK, "failed to initialize hardware");
+  params.comp_lvl = compression_level;
+  params.hw_buff_sz = hardware_buffer_size;
+  params.strm_buff_sz = stream_buffer_size;
+  params.input_sz_thrshold = input_size_threshold;
+
+  status = qzInit(&session_, params.sw_backup);
+  RELEASE_ASSERT(status == QZ_OK || status == QZ_DUPLICATE, "failed to initialize hardware");
+  status = qzSetupSession(&session_, &params);
+  RELEASE_ASSERT(status == QZ_OK || status == QZ_DUPLICATE, "failed to initialize hardware");
```
mythi: `failed to setup session` to distinguish from the error above?

envoy_qatzip/compressor/qatzip_compressor_impl.cc
```diff
@@ -0,0 +1,101 @@
...
+  auto status = qzCompressStream(&session_, &stream_, last);
+  // NOTE: stream_.in_sz and stream_.out_sz have changed their semantics after the call
+  //       to qzCompressStream(). Despite their name the new values are consumed input
+  //       and produced output (not available buffer sizes).
+  avail_out_ -= stream_.out_sz;
+  avail_in_ -= stream_.in_sz;
+  stream_.in = stream_.in + stream_.in_sz;
+  stream_.out = stream_.out + stream_.out_sz;
+  RELEASE_ASSERT(status == QZ_OK, "");
+  if (avail_out_ == 0) {
```
mythi, line 91: maybe add a comment here saying that the chunk is full and we copy the results?

envoy_qatzip/compressor/qatzip_compressor_impl.cc
```diff
@@ -0,0 +1,101 @@
...
+  int status = qzGetDefaults(&params);
+  RELEASE_ASSERT(status == QZ_OK, "failed to initialize hardware");
+  params.comp_lvl = compression_level;
+  params.hw_buff_sz = hardware_buffer_size;
+  params.strm_buff_sz = stream_buffer_size;
+  params.input_sz_thrshold = input_size_threshold;
+
+  status = qzInit(&session_, params.sw_backup);
+  RELEASE_ASSERT(status == QZ_OK || status == QZ_DUPLICATE, "failed to initialize hardware");
+  status = qzSetupSession(&session_, &params);
```
mythi, line 44: have you tried two filter chains with different settings?

envoy_qatzip/qatzip.proto
```diff
@@ -0,0 +1,44 @@
...
+
+package qatzip;
+
+import "envoy/extensions/filters/http/compressor/v3/compressor.proto";
+
+import "google/protobuf/wrappers.proto";
+
+import "validate/validate.proto";
+
+message Qatzip {
```
mythi, line 11: your sample config in this PR would be nice

README.md
```diff
@@ -13,12 +13,12 @@ Clone this repository with submodules:
...
     $ cd kubernetes-qat-envoy
-    $ wget https://01.org/sites/default/files/downloads/qat1.7.l.4.6.0-00025.tar.gz
+    $ wget https://01.org/sites/default/files/downloads/qat1.7.l.4.8.0-00005.tar.gz
 
 Check that the correct archive has been loaded by calculating its sha256 checksum:
 
-    $ sha256sum qat1.7.l.4.6.0-00025.tar.gz
-    8381567a11766ab89e556a41aad9a71031209f68b27ae0c49ff59757661162f4  qat1.7.l.4.6.0-00025.tar.gz
+    $ sha256sum qat1.7.l.4.8.0-00005.tar.gz
+    5d0770d20fd77c516a247f2917e47048432cea1394c00d9aa8b0b6a85030fd2c  qat1.7.l.4.8.0-00005.tar.gz
```
mythi: this is automatically done by the scripts/Dockerfile so perhaps just drop it? Also, there's a link to `qat1.7.l.4.6.0-00025`  in the introduction.

Dockerfile.envoy
```diff
@@ -1,9 +1,6 @@
 FROM debian:sid as builder
```
mythi, line 1: we should drop this and only have `Dockerfile.openssl` and `Dockerfile.boringssl`  that are clearlinux based.

offtopic: we should add Github actions build matrix for `openssl` and `boringssl`.

mythi (changes requested):
very nice! just a few asks

rojkov:
I've updated the submodules and added an example config for QATzip. But the config format is going to be changed radically when I have https://github.com/rojkov/envoy/commit/2fe000375d7221ff5d9d2c2059dc350eecc7df1b merged upstream (Matt wants to deprecate separate compression filters and to have only one configurable with different compression libs).
I'll update this PR after I finish polishing the config format commit.

envoy_qatzip/compressor/qatzip_compressor_impl.cc
rojkov: No, I haven't. I assumed `QzSession_T` contains state data specific to a session(stream). I check in QATzip's source code it's not global and multiple sessions with different params can coexist at the same time. Then QAT handles chunks of data with the session params included in the chunks, so it's stateless.

But yeah, would be nice to test it, not to assume...

Dockerfile.envoy
rojkov: > we should drop this and only have `Dockerfile.openssl` and `Dockerfile.boringssl` that are clearlinux based.

I'd rather do it in a followup PR.

envoy_qatzip/compressor/qatzip_compressor_impl.cc
mythi: +1

Dockerfile.envoy
mythi: yes makes sense

mythi (approved):

PR #45, opened 2020-05-12
images: openssl: drop clang-9 workaround

mythi:


ipuustin (approved):
Looks good, also the `--no-boot-update' flag.

rojkov (approved):

PR #47, opened 2020-05-25
update build instructions and configs

mythi:
Some of the content had gotten outdated and the setup did not work
anymore. Update the relevant pieces.

### DIFF
--- a/deployments/nginx-behind-envoy-deployment-no-qat.yaml
+++ b/deployments/nginx-behind-envoy-deployment-no-qat.yaml
@@ -39,7 +39,6 @@ data:
               stat_prefix: ingress_http
               route_config:
                 name: local_route
-                require_ssl: all
                 virtual_hosts:
                 - name: backend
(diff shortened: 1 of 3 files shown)


deployments/nginx-behind-envoy-deployment-no-qat.yaml
```diff
@@ -39,7 +39,6 @@ data:
               stat_prefix: ingress_http
               route_config:
                 name: local_route
-                require_ssl: all
```
ipuustin, line 42: Do you think this should be replaced with `require_tls` or removed completely?

mythi: I thought about `require_tls` but decided to drop this completely. To me it looks this is a `route.VirtualHost` message anyways so it's not in the right place and, also, our `return 200` config does not specify this option...

ipuustin (approved):

PR #50, opened 2020-07-20
QATzip fixes

rojkov:
1. Create one QATzip session per worker thread.

2. Do not rely on QATzip for wrapping compressed streams with correct headers and trailers. Instead set up QATzip to disable this wrapping with ```params.data_fmt = QZ_DEFLATE_RAW;``` (anyway it wraps not stream, but  every output buffer) and wrap the stream right in the filter instead. This way in-browser decompression should work.

### DIFF
--- a/envoy_qatzip/compressor/qatzip_compressor_impl.cc
+++ b/envoy_qatzip/compressor/qatzip_compressor_impl.cc
@@ -9,43 +9,18 @@
 namespace Envoy {
 namespace Compressor {
 
-QatzipCompressorImpl::QatzipCompressorImpl() : QatzipCompressorImpl(4096) {}
+QatzipCompressorImpl::QatzipCompressorImpl(QzSession_T* session)
+    : QatzipCompressorImpl(session, 4096) {}
 
-QatzipCompressorImpl::QatzipCompressorImpl(size_t chunk_size)
-    : initialized_{false}, chunk_size_{chunk_size}, avail_out_{chunk_size},
-      chunk_char_ptr_(new unsigned char[chunk_size]), session_{}, stream_{} {
+QatzipCompressorImpl::QatzipCompressorImpl(QzSession_T* session, size_t chunk_size)
+    : chunk_size_{chunk_size}, avail_out_{chunk_size},
+      chunk_char_ptr_(new unsigned char[chunk_size]), session_{session}, stream_{} {
+  RELEASE_ASSERT(session_ != nullptr,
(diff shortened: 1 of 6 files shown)


envoy_qatzip/compressor/qatzip_compressor_impl.cc
```diff
@@ -9,43 +9,20 @@
...
-  if (initialized_) {
-    qzEndStream(&session_, &stream_);
-    qzTeardownSession(&session_);
-    qzClose(&session_);
-  }
+QatzipCompressorImpl::QatzipCompressorImpl(QzSession_T* session)
+    : QatzipCompressorImpl(session, 4096) {}
+
+QatzipCompressorImpl::QatzipCompressorImpl(QzSession_T* session, size_t chunk_size)
+    : chunk_size_{chunk_size}, avail_out_{chunk_size-10},
```
ipuustin, line 17: I guess `10` is due to the header size? What values can `chunk_size` have?

envoy_qatzip/qatzip_filter.cc
```diff
@@ -54,10 +65,32 @@ unsigned int QatzipFilterConfig::streamBufferSizeUint(Protobuf::uint32 stream_bu
...
+    qzClose(&session_);
+  }
+}
+
+QzSession_T* QatzipFilterConfig::QatzipThreadLocal::getSession() {
+  // The session must be initialized only once in every worker thread.
+  if (!initialized_) {
+
+    int status = qzInit(&session_, params_.sw_backup);
+    RELEASE_ASSERT(status == QZ_OK || status == QZ_DUPLICATE, "failed to initialize hardware");
```
ipuustin, line 87: If the session must be initialized only once, isn't `QZ_DUPLICATE` an error case?

ipuustin (commented):
Looks good, just a few questions!

envoy_qatzip/qatzip_filter.cc
mythi: The API doc says it's not an error but only a sign that the function was not processed

envoy_qatzip/compressor/qatzip_compressor_impl.cc
rojkov: Currently `chunk_size` is a constant `4096`. Later it's going to be configured through a config.
I've added a TODO notice not to forget to add a boundary check to the config definition.

PR #52, opened 2020-08-06
Update Envoy+OpenSSL Dockerfile

mythi:

### DIFF
--- a/Dockerfile.openssl
+++ b/Dockerfile.openssl
@@ -1,8 +1,8 @@
 FROM clearlinux:base as builder
 
-ARG QAT_DRIVER_RELEASE="qat1.7.l.4.9.0-00008"
-ARG QAT_ENGINE_VERSION="v0.5.44"
-ARG QAT_ENGINE_FLAGS="--disable-multibuff-offload"
+ARG QAT_DRIVER_RELEASE="qat1.7.l.4.10.0-00014"
+ARG QAT_ENGINE_VERSION="v0.5.46"
+ARG IPSEC_MB_VERSION="v0.54"
 
 RUN swupd update && \
     swupd bundle-add \
@@ -18,6 +18,7 @@ RUN swupd update && \
     os-core-dev && \
     git clone -b $QAT_ENGINE_VERSION https://github.com/intel/QAT_Engine && \
     git clone https://github.com/intel/ipp-crypto && \
+    git clone -b $IPSEC_MB_VERSION https://github.com/intel/intel-ipsec-mb && \
     mkdir QAT_Lib && cd QAT_Lib && \
     curl -L https://01.org/sites/default/files/downloads/$QAT_DRIVER_RELEASE.tar.gz | tar zx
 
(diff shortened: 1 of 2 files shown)


Dockerfile.openssl
```diff
@@ -58,15 +58,28 @@ RUN sed -i -e '/OPENSSL_USE_STATIC_LIBS/d' CMakeLists.txt && \
...
+RUN make && make install LIB_INSTALL_DIR=/usr/lib64
+
 WORKDIR /QAT_Engine
 RUN ./autogen.sh && \
     ./configure \
-    $QAT_ENGINE_FLAGS \
+    --enable-ipsec_offload \
+    --enable-multibuff_offload \
+    --with-openssl_install_dir=/usr/lib64 \
+    --with-multibuff_dir=/ipp-crypto/sources/ippcp/crypto_mb \
```
mythi: I learnt this is redundant/obsolete. I'll drop it

mythi: fixed

Dockerfile.openssl
```diff
@@ -58,15 +58,27 @@ RUN sed -i -e '/OPENSSL_USE_STATIC_LIBS/d' CMakeLists.txt && \
...
 WORKDIR /QAT_Engine
 RUN ./autogen.sh && \
     ./configure \
-    $QAT_ENGINE_FLAGS \
+    --enable-ipsec_offload \
+    --enable-multibuff_offload \
+    --with-openssl_install_dir=/usr/lib64 \
+    --with-multibuff_install_dir=/usr/local \
+    --enable-openssl_install_build_arch_path && \
+    sed -i -e s:\"qat\":\"avx512\": e_qat.c && \
```
ipuustin: Do I understand it right that you are building the engine two times (`avx512.so` and `qat.so`)? I'm just thinking that this `sed` call looks bit fragile for future QAT engine releases. Is there no way to configure the engine run-time to use a particular acceleration method?

ipuustin (commented):
Looks good, just one comment/question...

Dockerfile.openssl
mythi: Currently there's no way to select the acceleration method runtime for rsa mb vs rsa QAT and either one must be select built time. With this PR, I wanted to split the functionality to qat-only and avx512-only and provide users configuration options outside. I can make the `sed` filter more robust...

ipuustin: There's `DISABLE_QAT_OFFLOAD` flag for the engine config, but maybe that's indeed not enough.

mythi: Right. Even if you specify "enable QAT" and "enable RSA MB" config options build-time, only the former is enabled if QAT lib is found from the build env: https://github.com/intel/QAT_Engine/blob/master/.tools/configure.ac#L244-L255

Dockerfile.openssl
```diff
@@ -1,8 +1,8 @@
 FROM clearlinux:base as builder
 
-ARG QAT_DRIVER_RELEASE="qat1.7.l.4.9.0-00008"
-ARG QAT_ENGINE_VERSION="v0.5.44"
-ARG QAT_ENGINE_FLAGS="--disable-multibuff-offload"
+ARG QAT_DRIVER_RELEASE="qat1.7.l.4.10.0-00014"
```
ipuustin, line 3: Please update version also to the README (even though I tested that userspace 4.10 appears to work fine with driver from 4.9).

Dockerfile.openssl
ipuustin: Alright. I filed an issue to QAT engine (https://github.com/intel/QAT_Engine/issues/158).

Dockerfile.openssl
mythi: updated

Dockerfile.openssl
mythi: I updated the `sed` magic.

ipuustin (approved):
LGTM

PR #56, opened 2020-10-13
Dockerfile.openssl: fix ipp-crypto build

mythi:
ipp-crypto repository did not provide a tag with the necessary changes
included until recently so we were using the master branch.

Now the tag was added (and master broke our build) so we move to use it.


rojkov (approved):
LGTM!

PR #57, opened 2020-10-27
Add qatzip compressor plugin

ozhuraki:
- Add initial version of qatzip compressor plugin
- Simplify trailer addition
- Add an example configuration
- Remove obsolete plugin
- Add an example BUILD file for QAT libraries


rojkov (approved):
LGTM!
I defer to @mythi or @ipuustin for merging.

examples/qat.BUILD
```diff
@@ -0,0 +1,57 @@
...
+	"cp -v $BUILD_TMPDIR/quickassist/utilities/libusdm_drv/*.h $BUILD_TMPDIR/$INSTALL_PREFIX/include/",
+],
+    static_libraries = [
+        "libqat.a",
+	"libusdm_drv.a",
+        "libosal.a",
+        "libadf.a",
+    ],
+    out_include_dir = "include",
+    linkopts = ["-ludev"],
```
ipuustin, line 56: How is the `-ludev` dependency handled? Does libudev need to present in the build system? What about during runtime?

examples/qat.BUILD
```diff
@@ -0,0 +1,57 @@
...
+	"cp -v $BUILD_TMPDIR/build/*.a $BUILD_TMPDIR/$INSTALL_PREFIX/lib/",
+        "mkdir -p $BUILD_TMPDIR/$INSTALL_PREFIX/include",
+	"cp -v $BUILD_TMPDIR/quickassist/include/*.h $BUILD_TMPDIR/$INSTALL_PREFIX/include/",
+	"cp -v $BUILD_TMPDIR/quickassist/include/dc/*.h $BUILD_TMPDIR/$INSTALL_PREFIX/include/",
+	"cp -v $BUILD_TMPDIR/quickassist/lookaside/access_layer/include/*.h $BUILD_TMPDIR/$INSTALL_PREFIX/include/",
+	"cp -v $BUILD_TMPDIR/quickassist/utilities/libusdm_drv/*.h $BUILD_TMPDIR/$INSTALL_PREFIX/include/",
+],
+    static_libraries = [
+        "libqat.a",
+	"libusdm_drv.a",
```
ipuustin, line 51: Nit: indentation

examples/qat.BUILD
```diff
@@ -0,0 +1,57 @@
...
+    make_commands = [
+    	"[[ \
+           ! -f $BUILD_TMPDIR/build/libqat.a || \
+	   ! -f $BUILD_TMPDIR/build/libusdm_drv.a  || \
+	   ! -f $BUILD_TMPDIR/build/libosal.a || \
+	   ! -f $BUILD_TMPDIR/build/libadf.a \
+	 ]] && make quickassist-all",
+	"mkdir -p $BUILD_TMPDIR/$INSTALL_PREFIX/lib",
+	"cp -v $BUILD_TMPDIR/build/*.a $BUILD_TMPDIR/$INSTALL_PREFIX/lib/",
+        "mkdir -p $BUILD_TMPDIR/$INSTALL_PREFIX/include",
```
ipuustin, line 43: Nit: indentation

examples/qatzip-envoy-conf.yaml
```diff
@@ -39,11 +39,17 @@ static_resources:
...
                   cluster: cluster_0
           http_filters:
-          - name: envoy.qatzip
-            config:
-              compressor:
-                content_length: 48
+          - name: envoy.filters.http.compressor
+            typed_config:
+              "@type": type.googleapis.com/envoy.extensions.filters.http.compressor.v3.Compressor
+              content_length: 10
```
ipuustin, line 48: How did you choose 10 over the previous value (48)? Just curious if there's a perf benefit.

examples/qatzip-envoy-conf.yaml
```diff
@@ -39,11 +39,17 @@ static_resources:
...
+              compressor_library:
+                name: testqat
+                typed_config:
+                  "@type": type.googleapis.com/qatzip.compressor.Qatzip
+                  hardware_buffer_size: sz_8k
```
ipuustin, line 53: Is `sz_8k` a kernel type? Do we want to use it in Envoy config?

qatzip/compressor/BUILD
```diff
@@ -0,0 +1,42 @@
...
+
+envoy_cc_extension(
+    name = "config",
+    srcs = ["config.cc"],
+    hdrs = ["config.h"],
+    security_posture = "robust_to_untrusted_downstream",
+    repository = "@envoy",
+    deps = [
+        ":compressor_lib",
+		":qatzip_cc_proto",
```
ipuustin, line 37: Nit: indentation. Envoy has a Bazel file check, run it to get consistent Bazel file look. The files are then compatible with upstream also.

examples/qat.BUILD
ozhuraki: @ipuustin 
My undestanding is that libudev needs to be present at the link tie and also at runtime.

qatzip/compressor/BUILD
ozhuraki: @ipuustin 

OK, thanks, will check and update these.

examples/qatzip-envoy-conf.yaml
rojkov: Why not?

ipuustin: Maybe put the question in other way: is the user expected to correspond this value to something in the kernel or libqat? Or is string `sz_8k` mentioned in the HW docs? Otherwise it seems to me that something like "8k" would sound more natural. Or how the user expected to choose which value to put here? Sorry, I'm a newbie in the QAT compression world. :-)

rojkov: I see. QATZip expects integer here. Yet this integer can be of 6 possible values only. That's why I resorted to Enum. I guess I tried to name the enum values as 8k, 16k,.. first, but then realized that literals should start with a letter. So I decided to add the `sz_` prefix to indicate it's an Enum explicitly and not to encourage users to try something like 8M.

newblaque (approved):

rojkov:
Seems like this PR stalled, so I've created #60 which addresses the review comments and can be merged instead of this one or after.

PR #58, opened 2020-10-30
Dockerfile.openssl improvements

ozhuraki:
- Minor readability updates
- Group QAT library, IPSEC_MB, IPP_CRYPTO, QAT_ENGINE statements
- Drop extra compilation options in bazel rc files
- Move envoy compilation into the own stage
- Update llvm to llvm10


Dockerfile.openssl
```diff
@@ -75,9 +75,6 @@ WORKDIR /
 COPY . .
 
 ENV HOME /root
-RUN eval $(cut -d = -f 2- < clang.bazelrc) && \
```
mythi, line 40: this was part of locking to clang9. squash with that removal commit?

Dockerfile.openssl
```diff
@@ -19,13 +19,12 @@ RUN swupd update && \
     os-core-dev && \
     git clone -b $QAT_ENGINE_VERSION https://github.com/intel/QAT_Engine && \
     git clone -b $IPP_CRYPTO_VERSION https://github.com/intel/ipp-crypto && \
-    git clone -b $IPSEC_MB_VERSION https://github.com/intel/intel-ipsec-mb && \
```
mythi, line 22: we had everything downloaded in one stage because if the build fails the source comes from cache.

Dockerfile.openssl
```diff
@@ -70,15 +70,12 @@ RUN git clone -b $QAT_ENGINE_VERSION https://github.com/intel/QAT_Engine && \
     make && \
     make install
 
-WORKDIR /
```
mythi: what is the reason for this commit? `WORKDIR /envoy`?

The bigger problem in our build is `COPY . .` because it trashes the cache when the repo changes...

Dockerfile.openssl
```diff
@@ -28,19 +28,6 @@ RUN mkdir QAT_Lib && \
     install -m 755 build/libusdm_drv_s.so /usr/lib/ && \
     install -m 755 build/adf_ctl /usr/bin/
 
-WORKDIR /
```
mythi, line 35: the current ordering was made on purpose (it was the ipp-crypto/ipsec-mb changing frequently) but I guess things have settled and this change makes sense. 

Perhaps move this commit later in the series because you're going to modify some of these lines later still

mythi (commented):
Nice improvements @ozhuraki, thanks!

Dockerfile.openssl
ozhuraki: @mythi 

Thanks! Makes sense and done.

Dockerfile.openssl
ozhuraki: @mythi 
> what is the reason for this commit? WORKDIR /envoy?

The logic here is that copying everything recursively under “/” might accidentally clash with the distribution’s directories/files.

It doesn’t harm and makes it easier to understand what comes from where.

Also from the point of view of keeping the build output under /build, since bazel doesn’t support when the build output is under the workspace.

Would it make sense?

Dockerfile.openssl
ozhuraki: @mythi 
> Perhaps move this commit later in the series because you're going to modify some of these lines later still

Makes sense. OK, moved those lines into 0eca540, please take a look.

Thanks for the review!

Dockerfile.openssl
mythi: Makes sense but how about making it `/envoy-build` to be explicit that it's separate from envoy submodule?

ozhuraki: @mythi 
> Makes sense but how about making it /envoy-build to be explicit that it's separate from envoy submodule?

Would be fine, except in connection with storing the build output under /build might be slightly confusing.

Maybe envoy-qat?

.bazelrc
```diff
@@ -1,6 +1,3 @@
 import %workspace%/envoy/.bazelrc
 
-build --copt="-DENVOY_SSL_VERSION=\"OpenSSL\""
-build --cxxopt=-Wno-error=old-style-cast
-build --define=boringssl=disabled
```
rojkov, line 5: Why did you decide to remove this line and `ENVOY_SSL_VERSION`?

ozhuraki: @rojkov 

Thanks for the input here. It slipped my attention, let's keep it.

ozhuraki: @rojkov 

Fixed. Please take a look.

build_envoy.sh
```diff
@@ -0,0 +1,29 @@
+#!/usr/bin/env bash
+
+mkdir -p build
+
+ENVOY_DEV_IMAGE=${1}
```
ipuustin: Maybe provide a default value and override it with the argument?

build_envoy.sh
```diff
@@ -0,0 +1,29 @@
+#!/usr/bin/env bash
+
+mkdir -p build
```
ipuustin: Would this file fit better in `scripts` directory? You would need to set the build dir differently then.

build_envoy.sh
```diff
@@ -0,0 +1,29 @@
...
+OPTIND=2
+while getopts "i" opt; do
+    case $opt in
+        i) CMD="" ;;
+    esac
+done
+
+docker run \
+       --rm \
+       -it \
```
ipuustin: Is `-i` needed because this is non-interactive?

Dockerfile.openssl
```diff
@@ -6,82 +6,79 @@ ARG IPSEC_MB_VERSION="v0.54"
 ARG IPP_CRYPTO_VERSION="ippcp_2020u2"
```
ipuustin, line 7: While you are updating the Dockerfile, try bumping up the QAT library version to `qat1.7.l.4.11.0-00001`. Are the other components the latest versions?

.bazelrc
rojkov: I'd preserve `build --define=boringssl=disabled` to explicitly disable all the extensions that depend on BorinSSL.

ozhuraki: @rojkov 
> I'd preserve build --define=boringssl=disabled to explicitly disable all the extensions that depend on BorinSSL.

Let's keep it then. Done.

Dockerfile.openssl
ozhuraki: @ipuustin 
> While you are updating the Dockerfile, try bumping up the QAT library version to qat1.7.l.4.11.0-00001.

Done.

> Are the other components the latest versions?

There are newer versions of IPP_CRYPTO, IPSEC_MB, QAT_ENGINE_VERSION.

IPSEC_MB, QAT builds/installs, IPP_CRYPTO, QAT_ENGINE fail, so, perhaps, let sort these updates later one by one.

build_envoy.sh
ozhuraki: @ipuustin 

> Would this file fit better in scripts directory?

OK, let's move it there.

 > You would need to set the build dir differently then.

Updated to "mkdir -p ${PWD}/build". Not sure if it makes more sense this way, please take a look and comment.

ipuustin: The other script in the `scripts` dir does gets the directory paths this way:

```
SCRIPT_DIR="$(dirname "$(realpath "${BASH_SOURCE[0]}")")"
BASE_DIR="$(realpath "${SCRIPT_DIR}/..")"
```

Your command would be `mkdir -p "${BASE_DIR}/build"` then.

Dockerfile.openssl
mythi: I can take a look `ipp-crypto_2020_update3`.

build_envoy.sh
ozhuraki: @ipuustin 

Thanks. OK, added & updated.

build_envoy.sh
ozhuraki: @ipuustin 

Removing "-i" (detaching from STDIN) makes it cumbersome to interrupt it.

So, pehaps, let's keep it as a convenience?

ipuustin: Fine with me.

Dockerfile.openssl
mythi: @ozhuraki For ipp-crypto and QAT Engine, follow these changes: https://github.com/intel/intel-device-plugins-for-kubernetes/pull/492

Dockerfile.openssl
mythi: +1

ozhuraki: @mythi 
OK, done.

Dockerfile.openssl
ozhuraki: @mythi 
> For ipp-crypto and QAT Engine [...]

Done.

After QAT Engine update I don't see libqat.so anymore, but there's libqatengine.so. I included it into the final image.

Is this OK?

ipuustin: The openssl config file should point to the right engine. You need to try out if things still work (maybe just by using `openssl speed` or similar).

mythi: @ozhuraki the new engines are called `qat-sw`  and `qat-hw`. `libqatengine.so` should not be added

ozhuraki: @mythi 

Thanks for the clarification. Removed it.

Dockerfile.openssl
```diff
@@ -1,101 +1,87 @@
...
+    swupd bundle-add --skip-diskspace-check --no-boot-update \
+        os-core-dev \
+        c-basic \
+        llvm10 \
+        curl \
+        git \
+        python3-basic \
+        devpkg-systemd \
+        devpkg-openssl && \
+    curl -fsSL --output /usr/bin/bazel https://github.com/bazelbuild/bazelisk/releases/download/v1.6.1/bazelisk-linux-amd64 && \
```
mythi: looks like bazelisk version could be easily added via an ARG setting

ozhuraki: @mythi 

OK, done.

mythi (approved):
+1

Dockerfile.openssl
ozhuraki: @ipuustin 
> The openssl config file should point to the right engine. You need to try out if things still work (maybe just by using openssl speed or similar).

I have checked, there's a problem with the new version of QAT_ENGINE:

```
# openssl speed -engine=qat-hw
invalid engine "qat-hw"
139655177307456:error:25066067:DSO support routines:dlfcn_load:could not load the shared library:crypto/dso/dso_dlfcn.c:118:filename(/usr/lib64/engines-1.1/qat-hw.so): /usr/lib64/engines-1.1/qat-hw.so: undefined symbol: qat_num_instances
[...]
```

Could anyone take a look? Or shall we, perhaps, handle library updates separately?

mythi (approved):

Dockerfile.openssl
ipuustin: Let's handle this version update separately then.

ipuustin (approved):

PR #61, opened 2021-06-08
Update Envoy+BoringSSL+QAT to the latest version

derekguo001:
Update Envoy to the latest version, and the configuration files
also need to be updated.

### DIFF
--- a/Dockerfile.boringssl
+++ b/Dockerfile.boringssl
@@ -23,20 +23,21 @@ RUN mkdir /install_root \
     --no-boot-update \
     && rm -rf /install_root/var/lib/swupd/*
 
-RUN swupd bundle-add os-core-dev c-basic python3-basic
+# GCC11 does not work with Envoy and we need to use GCC10.
+RUN swupd bundle-add os-core-dev c-extras-gcc10 python3-basic
 
 # for installing bazel as user
 ENV HOME /root
 
 # bazel 3.4 is not packaged to Clear Linux yet, change to that bundle when it is
-RUN curl -LO https://github.com/bazelbuild/bazel/releases/download/3.4.1/bazel-3.4.1-installer-linux-x86_64.sh
-RUN chmod +x bazel-3.4.1-installer-linux-x86_64.sh
-RUN ./bazel-3.4.1-installer-linux-x86_64.sh --user
+RUN curl -LO https://github.com/bazelbuild/bazel/releases/download/4.1.0/bazel-4.1.0-installer-linux-x86_64.sh
+RUN chmod +x bazel-4.1.0-installer-linux-x86_64.sh
+RUN ./bazel-4.1.0-installer-linux-x86_64.sh --user
 
-RUN git clone --recurse-submodules --single-branch --branch v1.15.0-qat1 https://github.com/ipuustin/envoy.git envoy
+RUN git clone --recurse-submodules --single-branch --branch qat-provider-main-v5 https://github.com/ipuustin/envoy.git envoy
 
 # build and install Envoy
-RUN cd envoy && /root/.bazel/bin/bazel build --copt="-Wno-uninitialized" --copt="-Wno-maybe-uninitialized" --copt="-Wno-stringop-truncation" --verbose_failures -c opt -s //source/exe:envoy-static --extra_toolchains=@rules_python//python:autodetecting_toolchain_nonstrict \
+RUN cd envoy && CC=gcc-10 CXX=g++-10 /root/.bazel/bin/bazel build --copt="-Wno-uninitialized" --copt="-Wno-maybe-uninitialized" --copt="-Wno-stringop-truncation" --verbose_failures -c opt -s //source/exe:envoy-static --extra_toolchains=@rules_python//python:autodetecting_toolchain_nonstrict \
     && install -D /envoy/LICENSE /install_root/usr/local/share/package-licenses/envoy/LICENSE \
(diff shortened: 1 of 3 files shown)


Dockerfile.boringssl
```diff
@@ -5,11 +5,11 @@
 # CLEAR_LINUX_VERSION="--version=29970"
 #
 # This is used on release branches before tagging a stable version.
-ARG CLEAR_LINUX_BASE=clearlinux:base
+ARG CLEAR_LINUX_BASE=clearlinux@sha256:b8e5d3b2576eb6d868f8d52e401f678c873264d349e469637f98ee2adf7b33d4
 
 FROM ${CLEAR_LINUX_BASE} as builder
 
-ARG CLEAR_LINUX_VERSION=
+ARG CLEAR_LINUX_VERSION="--version=29970"
```
ipuustin: Hmm, what's special about this version?

derekguo001: Without the env variable, the build will fail.
In my opinion, the version of all components should be fixed.
This can avoid some compilation problems caused by new version in the future.

derekguo001: It might be better to submit these modifications in another patch.

ipuustin: I would rather find out why the build fails with the latest version -- we should be able to target the latest Clear Linux.

mythi: @derekguo001 the build has failures with GCC11. Can you update this PR so that the latest Clear Linux version is used but GCC10 is used (from `c-extras-gcc10` bundle)

derekguo001: @mythi Hi, Mikko, I tried c-extras-gcc10 bundle with clearlinux:base, but GCC10 could not build Envoy correctly.

The modification in Dockerfile.boringssl:
`RUN swupd bundle-add os-core-dev c-extras-gcc10 python3-basic`

Compile Error:
`
Use --sandbox_debug to see verbose messages from the sandbox
external/boringssl/src/crypto/curve25519/curve25519.c:503:57: error: argument 2 of type 'const uint8_t[32]' {aka 'const unsigned ch
ar[32]'} with mismatched bound [-Werror=array-parameter=]
  503 | int x25519_ge_frombytes_vartime(ge_p3 *h, const uint8_t s[32]) {
      |                                           ~~~~~~~~~~~~~~^~~~~
In file included from external/boringssl/src/crypto/curve25519/curve25519.c:33:
external/boringssl/src/crypto/curve25519/internal.h:109:58: note: previously declared as 'const uint8_t *' {aka 'const unsigned cha
r *'}
  109 | int x25519_ge_frombytes_vartime(ge_p3 *h, const uint8_t *s);
      |                                           ~~~~~~~~~~~~~~~^
external/boringssl/src/crypto/curve25519/curve25519.c:823:57: error: argument 2 of type 'const uint8_t *' {aka 'const unsigned char
 *'} declared as a pointer [-Werror=array-parameter=]
  823 | void x25519_ge_scalarmult_base(ge_p3 *h, const uint8_t *a) {
      |                                          ~~~~~~~~~~~~~~~^
In file included from external/boringssl/src/crypto/curve25519/curve25519.c:33:
external/boringssl/src/crypto/curve25519/internal.h:117:56: note: previously declared as an array 'const uint8_t[32]' {aka 'const u
nsigned char[32]'}
  117 | void x25519_ge_scalarmult_base(ge_p3 *h, const uint8_t a[32]);
      |                                          ~~~~~~~~~~~~~~^~~~~
cc1: all warnings being treated as errors
Target //source/exe:envoy-static failed to build
`

mythi: @derekguo001 

> The modification in Dockerfile.boringssl:
> `RUN swupd bundle-add os-core-dev c-extras-gcc10 python3-basic`

I don't think this alone is not enough. You'd need to set CC to be gcc10 for Envoy build.

derekguo001: Sorry. I replaced c-basic with c-extras-gcc10 and did not notice that os-core-dev also contained another gcc11.
It's ok now.

Dockerfile.boringssl
```diff
@@ -23,17 +23,19 @@ RUN mkdir /install_root \
     --no-boot-update \
     && rm -rf /install_root/var/lib/swupd/*
 
-RUN swupd bundle-add os-core-dev c-basic python3-basic
+# GCC11 does not work with Envoy and we need to use GCC10.
+RUN swupd bundle-add os-core-dev c-extras-gcc10 python3-basic
+RUN ln -sf /usr/bin/gcc-10 /usr/bin/gcc && ln -sf /usr/bin/g++-10 /usr/bin/g++
```
mythi: We can keep `c-basic` and do it without symlinks like this:


```RUN cd envoy && CC=gcc-10 CXX=g++-10 /root/.bazel/bin/bazel build --copt="-Wno-uninitialized" --copt="-Wno-maybe-uninitialized" --copt="-Wno-stringop-truncation" --verbose_failures -c opt -s //source/exe:envoy-static --extra_toolchains=@rules_python//python:autodetecting_toolchain_nonstrict \```

mythi (changes requested):
Let's just set `CC` and `CXX`.

Dockerfile.boringssl
derekguo001: In this case we do not need `c-basic` anymore. So we can replace it with `c-extras-gcc10`.
BTW, `os-core-dev` also contains `gcc11`.

mythi (approved):

PR #67, opened 2022-03-19
Modify listner, filter, cluster syntax to match with envoy API v3 sta…

zhqffrank:
…ndards

### DIFF
--- a/deployments/nginx-behind-envoy-deployment.yaml
+++ b/deployments/nginx-behind-envoy-deployment.yaml
@@ -27,20 +27,24 @@ data:
             address: 0.0.0.0
             port_value: 9000
         filter_chains:
-          tls_context:
-            common_tls_context:
+          transport_socket:
+            name: envoy.transport_sockets.tls
+            typed_config:
+             "@type": type.googleapis.com/envoy.extensions.transport_sockets.tls.v3.DownstreamTlsContext            
+             common_tls_context:
               tls_certificates:
                 certificate_chain: { "filename": "/etc/envoy/tls/tls.crt" }
                 private_key_provider:
                   provider_name: qat
                   typed_config:
                     "@type": "type.googleapis.com/envoy.extensions.private_key_providers.qat.v3alpha.QatPrivateKeyMethodConfig"
-                    section_name: ${QAT_SECTION_NAME}
-                    poll_delay: ${POLL_DELAY}
+                    section_name: SHIM 
+                    poll_delay: 0.005s
                     private_key: { filename: "/etc/envoy/tls/tls.key" }
           filters:
           - name: envoy.http_connection_manager
-            config:
+            typed_config:
+              "@type": type.googleapis.com/envoy.extensions.filters.network.http_connection_manager.v3.HttpConnectionManager  
               codec_type: auto
               stat_prefix: ingress_http
               route_config:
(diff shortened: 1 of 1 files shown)


zhqffrank:
Modify listner, filter, cluster syntax to match with envoy API v3 standards for file nginx-behind-envoy-deployment.yaml

deployments/nginx-behind-envoy-deployment.yaml
```diff
@@ -106,22 +116,18 @@ spec:
             capabilities:
               add: ["IPC_LOCK"]
           command:
-            - "/envoy-static"
+            - "/usr/local/bin/envoy-static"
           args:
             - "-c"
             - "/etc/envoy/config/envoy-conf.yaml"
             - "--cpuset-threads"
```
mythi, line 123: `entrypoint.sh`  gets used if we drop both `command` and `args`. That might be better approach now that we have only one image.

deployments/nginx-behind-envoy-deployment.yaml
```diff
@@ -106,22 +116,18 @@ spec:
...
-            - name: "POLL_DELAY"
-              value: "2000000"
           resources:
             limits:
-              qat.intel.com/generic: 1
+              qat.intel.com/cy2_dc2: 1
               cpu: 2
               memory: "2G"
             requests:
+              qat.intel.com/cy2_dc2: 1
```
mythi, line 130: `limits` is enough so this can be dropped.

deployments/nginx-behind-envoy-deployment.yaml
```diff
@@ -106,22 +116,18 @@ spec:
...
             - "--cpuset-threads"
-          env:
-            - name: "QAT_SECTION_NAME"
-              value: "SHIM"
-            - name: "POLL_DELAY"
-              value: "2000000"
           resources:
             limits:
-              qat.intel.com/generic: 1
+              qat.intel.com/cy2_dc2: 1
```
mythi, line 126: my preference is to keep `generic`. for the intel-qat-kernel-plugin resources we could create a kustomization overlay that patches the resource type.

mythi (changes requested):
thanks! let's drop the `qat.intel.com` resource specific changes and focus the envoy config/commandline fixes.

deployments/nginx-behind-envoy-deployment.yaml
zhqffrank: Agree

zhqffrank: Agree

mythi:
@zhqffrank no need to close but if you plan to force push with updates remember to re-open this PR first

zhqffrank:
Reopen for merge

deployments/nginx-behind-envoy-deployment.yaml
```diff
@@ -27,20 +27,24 @@ data:
...
+             common_tls_context:
               tls_certificates:
                 certificate_chain: { "filename": "/etc/envoy/tls/tls.crt" }
                 private_key_provider:
                   provider_name: qat
                   typed_config:
                     "@type": "type.googleapis.com/envoy.extensions.private_key_providers.qat.v3alpha.QatPrivateKeyMethodConfig"
-                    section_name: ${QAT_SECTION_NAME}
-                    poll_delay: ${POLL_DELAY}
+                    section_name: SHIM
```
ipuustin, line 41: I think that if we go with the entrypoint script, we can't hardcode these values. The values are coming to the container as environmental variables (from the QAT device plugin), and the entrypoint script substitutes them in the configuration file.

ipuustin (changes requested):

deployments/nginx-behind-envoy-deployment.yaml
zhqffrank: agree

